PR #110, opened 2026-07-09
perf: reduce graph export and dependency scan memory

matiasdaloia:
## Summary

- Stream large JSON report/export paths instead of buffering whole encoded payloads before writing.
- Add compact graph-fragment internal edge encoding while preserving edge data for ingest/reconstruction.
- Stream-build callgraph finding graphs and optimize hot lookup paths used during large scans.
- Prune dependency callgraph inputs to packages that can sit on a user-to-crypto dependency path when the resolver graph proves it; fall back conservatively when it cannot.
- Populate Maven dependency graph metadata best-effort with a bounded `dependency:tree` call so Java scans can use the same pruning path.
- Drop no-finding dependency reports from memory after caching and use a safer Java default for dependency scan workers.

## Benchmarks

bcprov-jdk15on 1.70 mining-faithful scan, with both graph fragment and callgraph exports:

| Metric | Result |
|---|---:|
| Elapsed | 129.457s |
| Max RSS | ~3.69 GiB |
| Fragment output | ~209 MB |
| Callgraph output | ~656 MB |
| Findings | 1,748 |
| Functions | 19,230 |
| Internal edges | 3,907,909 |
| Crypto entry points | 2,156 |

## Dependency scanning changes

`--scan-dependencies` is not recursive-exponential: dependencies are canonicalized and scanned once. The remaining expensive path is the combined callgraph build. This PR reduces that cost by using resolver graph proof to avoid parsing dependencies that cannot be on any root-to-crypto-dependency path, while preserving the old all-deps behavior when the graph is missing or incomplete.

For Maven, graph extraction is best-effort and bounded. If `dependency:tree` is slow, missing, or incomplete, the scan keeps the conservative pre-existing behavior.

Important data-quality guard: the dependency callgraph is still built even when dependency source scans produce zero direct findings, because API-boundary synthesis and dependency bridge paths can require the graph.

## Validation

- [x] `git diff --check`
- [x] `make lint`
- [x] `GOCACHE=/private/tmp/crypto-finder-bcprov-bench/cache go test ./internal/engine -run 'TestDependency(ScanWorkers|Scanner_(CollectPackageSets|ScanSingleDep|ScanDependenciesParallel|ScanWithDependencies_NoDepsAndErrors))|TestMavenResolver'`
- [x] `GOCACHE=/private/tmp/crypto-finder-bcprov-bench/cache go test ./internal/dependency ./internal/cli ./internal/scan ./pkg/...`
- [x] Fresh-cache smoke: `client-encryption-java` with `--scan-dependencies --dep-ecosystem java --dep-workers=1 --no-cache --no-remote-rules`, isolated `HOME=/private/tmp/crypto-finder-client-java-smoke/run2/home`; exit 0, 47s, 84 assets, fragment exported.

## Notes

Full `go test ./...` was not run locally because the broader suite includes Docker/testcontainers-backed Postgres tests that are not suitable for this sandboxed worktree.

## Summary by CodeRabbit

* **New Features**
  * Graph-fragment exports now use a newer schema and include a compact internal-edge format for smaller output files.
  * Maven dependency scanning now captures dependency graph data when available.
  * Scan tooling now supports disabling findings caching entirely.

* **Bug Fixes**
  * JSON exports now stream directly to files, reducing memory use and improving reliability on large outputs.
  * Exported JSON preserves special characters without HTML escaping.

* **Performance**
  * Dependency and graph export handling is more efficient for large projects.

### DIFF
--- a/internal/callgraph/builder.go
+++ b/internal/callgraph/builder.go
@@ -92,9 +92,10 @@ func (b *Builder) PackageSeparator() string {
 func (b *Builder) BuildFromDirectories(packages, typeOnlyPackages []PackageDir) (*CallGraph, error) {
 	buildStart := time.Now()
 	graph := &CallGraph{
-		Functions:       make(map[string]*FunctionDecl),
-		Callers:         make(map[string][]string),
-		EdgeResolutions: make(map[string]EdgeResolution),
+		Functions:             make(map[string]*FunctionDecl),
+		Callers:               make(map[string][]string),
+		EdgeResolutions:       make(map[string]EdgeResolution),
+		EdgeResolutionsByPair: make(map[string][]EdgeResolution),
 	}
 
 	// Phase 1: Parse source files only for packages that need full analysis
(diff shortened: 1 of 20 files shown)


matiasdaloia:
## CodeRabbit Fixes Applied

Fixed 4 file(s) based on 3 unresolved CodeRabbit review thread(s).

**Files modified:**
- `internal/output/json.go`
- `internal/scan/fragment_export.go`
- `pkg/graphfrag/ingest.go`
- `pkg/graphfrag/ingest_test.go`

**Commit:** `23f93596`

**Validation:**
- `make lint` → 0 issues
- `GOCACHE=/private/tmp/crypto-finder-bcprov-bench/cache go test ./pkg/... ./internal/output ./internal/scan` → passing

PR #132, opened 2026-07-20
fix(graphfrag): preserve supporting parameter roles

### LINKED ISSUE #130: fix(graphfrag): preserve parameter roles on supporting calls
## Parent

#115

## Problem

Graph-fragment decoding preserves parameter roles on a supporting call's nested crypto call, but the public callgraph projection drops those roles. Consumers therefore receive the canonical call and its parameters without the derivation contract that was present in the source fragment.

This is observable through the public graph-fragment API: a role-bearing fragment becomes a role-free `supporting_calls` record after stitching and callgraph export.

## What to build

Preserve existing parameter-role records when a decoded graph fragment is stitched and projected into the public callgraph response. Reuse the existing parameter-role conversion used by crypto entry points rather than introducing another representation.

## Acceptance criteria

- [ ] A supporting call with parameter roles retains the same index, name, role, property, and derivation after graph-fragment decode, stitch, and callgraph export.
- [ ] The regression test exercises the public `DecodeFragment → StitchWithOptions → ToCallgraphExport` seam.
- [ ] Supporting calls without parameter roles remain unchanged and omit the field.
- [ ] Crypto-entry-point role projection remains unchanged.
- [ ] The current schema version remains valid because this restores an existing field rather than introducing a new one.
- [ ] `CHANGELOG.md` contains an Unreleased fix entry describing the restored public behavior.
- [ ] Focused graph-fragment tests and the full Go test suite pass.

## Blocked by

None — can start immediately.

matiasdaloia:
Closes #130

## Summary

- Preserve parameter roles when stitched supporting calls are projected to the public callgraph export.
- Keep role-free supporting calls unchanged and omit `parameter_roles`.
- Cover the full `DecodeFragment` → `StitchWithOptions` → `ToCallgraphExport` path.

## Root cause

`DecodeFragment` and stitching retained `CryptoCall.ParameterRoles`, but `exportCryptoCall` did not project them into `ExportCryptoCall`.

## Test plan

- [x] `go test ./pkg/graphfrag -run 'TestSupportingParameterRolesSurviveDecodeStitchAndExport|TestDecodeFragment_MapsRoleFields' -count=1`
- [x] `go test ./pkg/graphfrag/... -count=1`
- [x] `go test ./... -count=1`
- [x] `make lint`

## Self-review

- Standards review: no findings.
- Issue #130 spec review: no findings.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed stitched graph-fragment callgraph exports so supporting calls retain their parameter roles.
  * Preserved parameter role details through decoding, stitching, and export.
  * Ensured calls without defined parameter roles do not incorrectly include them in exports.
  * Maintained correct parameter roles for crypto entry points.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Scanner failure messages now explain documented semgrep/opengrep exit codes (`opengrep execution failed with exit code 7 (rule configuration contains no valid rules)`) and attach a sanitized stderr tail (ANSI-stripped, single line, capped on UTF-8 rune boundaries) to logs and error details; the failure debug log records rule configs + target instead of dumping the full command line. (#112)
 
 ### Fixed
+- Stitched graph-fragment callgraph exports now retain parameter roles on supporting calls. (#130)
 - Concurrent scans sharing the default rules cache no longer expose partial metadata or lose in-flight filtered rule files when another process refreshes the cache. (#128)
 - Rule files with zero rules (`rules: []`) are excluded from language filtering — previously treated as "unknown language", they always survived the filter and could become the sole config passed to opengrep, which exits with code 7 on an empty ruleset and hard-failed the whole scan. (#112)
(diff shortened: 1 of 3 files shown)


matiasdaloia:
Self-review complete at `728ced27c014bd8d06216173a14bbdf297fc6bd5`. The standards and issue-spec passes found no actionable problems. Focused graph-fragment tests, the full Go suite, and lint all pass locally.

PR #133, opened 2026-07-20
fix(scan): preserve supporting call overloads

### LINKED ISSUE #131: fix(graphfrag): make supporting-call overload selection deterministic
## Parent

#115

## Problem

Repeated graph-fragment exports for identical source, rules, and configuration can return different overloads in the supporting-call catalog while keeping the same record count. For example, a no-argument overload may be replaced by its integer overload, or a byte-array overload by its string overload.

This makes the semantic catalog unstable and prevents consumers from relying on complete canonical signatures to distinguish overloads.

## What to build

Make supporting-call catalog construction deterministic and overload-safe. Equivalent inputs must produce the same semantic set regardless of traversal or insertion order, and distinct overloads must not be collapsed using an identity weaker than their complete callable signature.

## Acceptance criteria

- [ ] Permuting otherwise equivalent supporting-call input order produces the same canonical-signature set.
- [ ] Distinct overloads remain distinguishable by declaring type, parameter types, and return type.
- [ ] Repeated exports of the same fixture produce the same semantic supporting-call catalog.
- [ ] Opaque supporting identifiers are not given a new cross-run stability guarantee; the assertion is on semantic records and their references.
- [ ] Existing finding-to-supporting-call references remain valid after deterministic catalog construction.
- [ ] `CHANGELOG.md` contains an Unreleased fix entry describing stable overload handling.
- [ ] Focused supporting-call tests and the full Go test suite pass.

## Blocked by

None — can start immediately. This is independent of #130.

matiasdaloia:
## Summary

- retain every declaration for a supporting-call method instead of selecting one from map traversal order
- match overload-decorated Java declarations by contract arity and deduplicate only by complete canonical signature
- cover repeated graph-fragment exports, overload semantics, and finding-to-catalog references

Closes #131

## Test plan

- [x] `go test ./internal/scan -run 'Test(BuildGraphFragmentExport_PreservesSupportingOverloadsAcrossRepeatedExports|DeriveContractSupportingCalls_PreservesOverloadsRegardlessOfInsertionOrder)$' -count=10`
- [x] `go test ./internal/scan -count=1`
- [x] `go test ./... -count=1`
- [x] `make lint`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Scanner failure messages now explain documented semgrep/opengrep exit codes (`opengrep execution failed with exit code 7 (rule configuration contains no valid rules)`) and attach a sanitized stderr tail (ANSI-stripped, single line, capped on UTF-8 rune boundaries) to logs and error details; the failure debug log records rule configs + target instead of dumping the full command line. (#112)
 
 ### Fixed
+- Supporting-call catalogs now preserve every callable overload deterministically instead of selecting one by traversal order. (#131)
 - Stitched graph-fragment callgraph exports now retain parameter roles on supporting calls. (#130)
 - Concurrent scans sharing the default rules cache no longer expose partial metadata or lose in-flight filtered rule files when another process refreshes the cache. (#128)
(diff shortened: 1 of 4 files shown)


matiasdaloia:
Self-review complete. The published diff stays within #131's scope: overload declarations are retained and arity-filtered, semantic deduplication uses the complete canonical signature, and supporting IDs remain opaque.

The focused regressions, `go test ./internal/scan`, the full Go suite, and `make lint` all pass. The graph-fragment boundary test also verifies every finding reference resolves to the emitted supporting-call catalog. No follow-up changes identified.

PR #243, opened 2026-08-13
fix(callgraph): restore java dependency call chains broken by two parser gaps

### LINKED ISSUE #244: Export findings with an explicit reachability field instead of inferring it from chain shape
## Problem

Whether a crypto finding is reachable from user code is currently expressed only by convention, and the convention is easy to misread.

A finding's `call_chains` today can mean three different things, and two of them look nearly identical:

| Situation | Export shape |
|---|---|
| User code calls crypto directly | one frame, **no** `dependency_info` |
| Chain runs from user code into a dependency | several frames, first one has **no** `dependency_info` |
| Nothing in user code reaches the crypto | one frame, **with** `dependency_info` |

The third case is not a failure — it is a valid negative answer — but it is produced by a fallback in `buildCallChains` (`internal/scan/export.go`) that synthesizes a single-node chain when `structuralTracebackChains` returns nothing:

```go
raw := structuralTracebackChains(ctx, containingFn)
if len(raw) == 0 {
    node := buildChainNode(ctx, containingFn.ID, containingFn.FilePath)
    chains = [][]callGraphChainNode{{node}}
}
```

So "no path exists" is serialized as a chain, and the only way to tell it apart from a real depth-1 chain is to inspect whether the first frame carries `dependency_info`. Every consumer has to know and reimplement that rule.

`callGraphExportFinding` has no field for this. `unresolved_reason` is not it: that field answers *which function contains this crypto call*, and is empty in the unreachable case because the attribution succeeded.

Concrete example from a Redis client sample (post #243): jedis's MD5 lives in `Hashing$1.hash(byte[])`, reachable only through jedis's sharding classes, which the example never uses. The finding is correct, attributed, and unreachable — but the export says that only implicitly.

## Proposal

Add an explicit tri-state field to `callGraphExportFinding`:

```go
UnresolvedReason string `json:"unresolved_reason,omitempty"` // which function contains it (existing)
Reachable        *bool  `json:"reachable,omitempty"`         // does user code reach it (new)
```

- `true` — at least one chain starts in a user package
- `false` — a containing function was found, but no chain reaches user code
- absent — the question does not apply

The third state matters and is easy to miss: in the mine path (`--export-graph-fragment`, no dependencies) `exportUserPackages` returns `nil` and chains run to graph roots. There is no notion of user code there, so `false` would be a lie — the field must be omitted.

Keep the two fields orthogonal. Folding a `"not_reachable"` value into `unresolved_reason` would mix a measurement failure with a valid negative result.

The name is not new vocabulary: the reachability API already exposes `reachable` on assets (`scanoss.api`, `openapi.yaml`). Client and server currently describe the same property in different ways.

## Considerations

- Bumps the partner-facing export schema (`pkg/graphfrag.CallgraphSchemaVersion`, currently `6.6`). Additive and backward compatible, but consumers should be told.
- Worth deciding alongside: what should the reachability API return for an unreachable finding? Today `entry_point_signatures` removes non-reachable assets entirely — *"they do NOT appear as `reachable=false` entries"*. If the full inventory with a reachability flag is wanted, that is a separate product decision.

## Why it matters

This is the distinction between *"this algorithm ships in my dependencies"* and *"my application executes this algorithm"* — the question that separates this tool from an ordinary SBOM. It should be a field, not a convention a consumer has to infer.

agustingroh:
closes #244 

## Problem

Java call chains for dependency crypto collapsed to a single frame, so the reachability half of the tool returned a constant answer: **no dependency finding ever traced back to user code.**

Two independent parser gaps caused it, both surfaced while validating a transitive-dependency sample.

### 1. Variables declared outside `local_variable_declaration` were never typed

The type collector only walked `local_variable_declaration` and `field_declaration`. A name bound by `try`-with-resources, enhanced-`for`, or `catch` never entered `varTypes`, so its receiver stayed untyped and the call fell through to the scanned file's own package:

```
jedis.set("k", "v")   →  com.example.jedis.set(?, ?)              (a callee in no graph)
                  want:  redis.clients.jedis.Jedis.set(String, String)
```

With no edge out of user code, `Tracer.classifyTraceItem` dropped every chain that failed to reach a user package, and `buildCallChains` fell back to emitting only the function containing the crypto call.

This shape is not exotic — it is how clients, connections and streams are idiomatically written. Three of five sample projects declare the crypto-carrying object this way (`redis-demo:80`, `postgres-demo:75`, `okhttp-demo:143`).

### 2. Anonymous class bodies were never registered as functions

The walk descended only into declared class and interface bodies, so methods inside `new Foo() { ... }` never became `FunctionDecl`s. A crypto call there belonged to no function and its finding exported with `unresolved_reason: "no_containing_function"` and zero chains — the shape jedis uses to publish its MD5 digest from an interface field.

## Approach

Variable binding keys on the grammar's own `type`/`name` fields rather than on a list of node types. The two multi-variable forms are excluded by the grammar itself (`local_variable_declaration` and `field_declaration` expose `type` but no `name`, since `int a = 1, b = 2;` binds several names), so they keep their existing handling and are not matched. Catch parameters need a dedicated path because the grammar nests their type inside `catch_type`.

Writing this as a list of node types is what hid the bug in the first place: enhanced-`for` was also broken and was not on any list until the field probe found it.

Two deliberate exclusions, both documented at the call site:

- **Multi-catch stays unbound.** The static type of `catch (IOException | SQLException e)` is the least upper bound of the alternatives, not any listed type. Binding the first would invent an edge into a concrete class the call may never reach, and a wrong edge is worse than a missing one — it can fabricate a crypto chain that does not exist.
- **`formal_parameter` stays with `parseJavaParameterList`**, which strips `final` and annotations and normalizes through `normalizeJavaReferenceType`. Binding it here as well would let raw field text overwrite that normalized form on the recursive descent.

Anonymous classes are registered under javac's binary name (`Owner$1`, `Owner$2`, numbered in source order) so the identity lines up with the class the bytecode index carries, and the instantiated supertype is recorded in `OwnerBases` so dispatch expansion can link interface call sites to the override. The pass runs for interfaces as well as classes: interface fields are implicitly `static final` and libraries use them to publish anonymous implementations.

## Measured effect — `redis-demo`

| | main | this branch |
|---|---|---|
| crypto entry points | 1 | **7** |
| findings with a call chain | 1 of 2 | **2 of 2** |
| `no_containing_function` | 1 | **0** |
| max chain depth | 1 | **6** |

```
javax.net.ssl.SSLSocketFactory.getDefault
  1. com.example.RedisTlsTransitive.demonstrateRedisTlsConnection(): void
  2. redis.clients.jedis.Jedis.get(String): String
  3. redis.clients.jedis.Connection.executeCommand(CommandArguments): Object
  4. redis.clients.jedis.Connection.sendCommand(CommandArguments): void
  5. redis.clients.jedis.Connection.connect(): void
  6. redis.clients.jedis.DefaultJedisSocketFactory.createSocket(): Socket
```

The second finding now reports a single frame at `redis.clients.jedis.util.Hashing$1.hash(byte[])`, which is the correct answer: `Hashing` is only referenced from jedis's sharding classes and the example never uses sharding, so that MD5 ships in the jar but is not reachable. Previously it could not be attributed to any function at all, so "unreachable" and "unlocated" were indistinguishable.

Earlier measurement of the variable-binding change alone across the other examples showed the same direction with no regressions: `gcp-sdk-demo` 137 → 158 entry points (max depth 1 → 12), `spring-kafka-demo` 13 → 34 (1 → 22), `postgres-demo` unchanged (its code calls no method on the resource variable), `okhttp-demo` unchanged at zero (no Kotlin ruleset exists, tracked separately).

## Tests

Two regression tests, both verified to fail without their fix.

`TestJavaParser_BindsSingleVariableDeclarationForms` covers try-with-resources, the `final` modifier, multi-resource, enhanced-`for` and single-type catch, and asserts multi-catch stays unbound. Without the fix:

```
set#2      resolved to com.example.jedis,  want redis.clients.jedis.Jedis
readLine#0 resolved to com.example.reader, want java.io.BufferedReader
doFinal#0  resolved to com.example.each,   want javax.crypto.Cipher
```

`TestJavaParser_RegistersAnonymousClassMethods` covers anonymous classes in an interface field initializer, a class field initializer and a method body, and asserts `OwnerBases` carries the instantiated supertype.

`make lint` reports 0 issues. `make test` passes except the Postgres testcontainers suite in `internal/engine`, which requires a running Docker daemon and fails identically on `main` in this environment.

## Notes

Additive only — no behavior removed, no schema change, no new flags. `docs/user-guide/` is untouched since no command, flag or output shape changed.

Follow-up filed separately: reachability is currently derivable only by convention (a chain whose first frame carries no `dependency_info` starts in user code), and deserves an explicit field.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Java callgraph accuracy for methods defined in anonymous and nested anonymous classes.
  - Added receiver type detection for try-with-resources, enhanced-for loops, and single-type catch blocks.
  - Prevented incorrect bindings for multi-catch variables.
  - Improved resolution across fluent varargs chains and partially modeled call paths.
  - Sanitized receiver text in exported results and added schema-aware fallback behavior.

---

## Review follow-up (24c382c)

`javaSingleDeclaredVar` now enumerates `resource` and `enhanced_for_statement` instead of accepting any node exposing `type`/`name` fields. `method_declaration` exposes both, and the recursive collectors reach method declarations inside anonymous class bodies, so the original rule let an anonymous `run()` rebind a local `Cipher run` to type `void` — a regression against `main`. Covered by `TestJavaParser_AnonymousClassScopeIsolation`.

Two review findings were reproduced and deliberately not acted on in this PR; both are explained in the review response comment and recorded in code:

- **Anonymous-class calls are also attributed to the enclosing function.** Real, but removing it regresses reachability today: the interface call site that should dispatch to the anonymous override does not resolve, because the callee is recorded under the simple type name (`CipherCallback`) while the declaration is registered under the nested one (`Wrapper.CipherCallback`). The over-attribution is currently the only path keeping callback crypto reachable.
- **Local classes are not traversed.** Pre-existing on `main`; closing it means registering local classes as types of their own.


agustingroh:
## Review follow-up

All three findings were reproduced against the branch before deciding on each. Two are fixed; one is deliberately deferred with evidence.

### Restrict `javaSingleDeclaredVar` to variable declarations — fixed in 24c382c

Confirmed, and it was a regression against `main`. `method_declaration` exposes `type` and `name` just like `resource` does, and the recursive collectors reach method declarations inside anonymous class bodies:

```java
Cipher run = null;
Runnable r = new Runnable() { @Override public void run() { } };
run.doFinal();
```

| | `run.doFinal()` resolves to |
|---|---|
| `main` | `javax.crypto.Cipher.doFinal#0` |
| this branch, before the fix | `com.example.void.doFinal#0` |

The binding rule now enumerates `resource` and `enhanced_for_statement`, with catch parameters keeping their own path. Reading the grammar's `type`/`name` fields is what makes those two uniform, but the fields alone do not mean "declares a variable" — the comment now says so, since the original wording claimed generality the rule does not have. Pinned by `TestJavaParser_AnonymousClassScopeIsolation`.

### Skip anonymous-class bodies during enclosing call extraction — not applied

The double-recording is real, and the semantics of the suggestion are right: constructing an object does not run its methods. But applying it regresses reachability, so it is deferred rather than dismissed.

Implementing it breaks `TestJavaParser_ResolvesAnonymousCallbackParameterReceiver`, which pins the over-attribution deliberately. Building the call graph for that fixture shows why — with the change applied the only surviving edge is `wrap → use`, and `cipher.init` becomes unreachable:

```java
byte[] wrap(SecretKey kek) {
    return use(kek, new CipherCallback() {
        public byte[] apply(Cipher cipher) { cipher.init(Cipher.WRAP_MODE, kek); }
    });
}
byte[] use(SecretKey kek, CipherCallback cb) { return cb.apply(null); }   // invoked immediately
```

The crypto genuinely executes when `wrap()` is called, so losing it is a false negative. The path that *should* carry it is broken by a separate defect — a nested-type naming mismatch:

```
declared:  com.example.crypto.(Wrapper.CipherCallback).apply#1   ownerType="interface"
call site: com.example.crypto.(CipherCallback).apply#1
```

No edge forms, so `expandInterfaceDispatch` never sees the call site and the anonymous override stays orphaned. The over-attribution is currently the only thing keeping callback crypto reachable — it compensates for that missing edge.

Correct order: fix nested-type callee naming, confirm dispatch reaches the anonymous override, then narrow `walkForCalls`. Doing the last step first trades a possible false positive for a guaranteed false negative, which is the worse direction for this tool. The rationale is recorded at the call site so the next reader does not repeat the analysis.

### Traverse local class declarations — confirmed, out of scope

Confirmed: a `class_declaration` inside a method body reaches neither `collectJavaClassDecls` (which iterates only direct children of a type body) nor `processClass`, so local classes and any anonymous class inside them stay unregistered.

This predates the PR — local classes were never registered on `main` either. Closing it means registering local classes as types of their own, with their own naming scheme, which is a capability rather than a fix and does not belong in this change. The misleading comment that claimed declared types "own their own subtree via processClass / processInterface" has been corrected to state the gap explicitly.

### Verification after the follow-up

`make lint` reports 0 issues; `internal/callgraph` and `internal/scan` pass. `redis-demo` is unchanged by the follow-up: 7 crypto entry points, 2 of 2 findings with a call chain, max depth 6.

PR #247, opened 2026-08-13
feat(scan): expose crypto reachability explicitly

### LINKED ISSUE #244: Export findings with an explicit reachability field instead of inferring it from chain shape

agustingroh:
Closes #244

## Problem

Whether user code reaches a crypto finding was only inferrable by convention, and an unreachable finding was published as an entry point beside the ones that actually run.

The tracer was already correct — `rootChainIsComplete` keeps only chains that reach a user package, so an unreachable finding produced an empty traceback. The exporter then synthesized a chain anyway:

```go
raw := structuralTracebackChains(ctx, containingFn)   // empty = nothing reaches it
if len(raw) == 0 {
    node := buildChainNode(ctx, containingFn.ID, containingFn.FilePath)
    chains = [][]callGraphChainNode{{node}}           // a chain is invented
}
```

`buildCryptoEntryPoints` then applies its rule — every function appearing in any chain is a potential entry point — and the invented chain puts the unreachable function in the list.

On a Redis client sample this published `redis.clients.jedis.util.Hashing$1.hash(byte[])` as a crypto entry point. That MD5 lives in jedis's `@Deprecated` sharding code; the example uses a plain client and no path from `main` reaches it. A CBOM built from that output claims the application uses MD5, which is false.

Consumers also had no field to read: `unresolved_reason` answers *which function contains this crypto*, and is empty in the unreachable case precisely because attribution succeeded. The reachability API already exposes `reachable` on assets, so client and server described the same property in different ways.

## The trap this avoids

The obvious fix — stop synthesizing the chain, or never index a chainless finding — would break mining.

On the mine path `exportUserPackages` returns `nil`, and a library's public API typically has no in-library callers. It is a graph root, `rootChainIsComplete` requires `len(chain) > 1` and fails, the traceback comes back empty, and **the synthesized chain is what makes it an entry point**. bcprov produces 2781 that way. Filtering unconditionally would delete exactly the entry points mining exists to produce.

So the meaning of an empty traceback depends on the mode, and the fix keys on that rather than on the chain's shape:

| Mode | `userPackages` | Empty traceback means |
|---|---|---|
| `--scan-dependencies` | `{RootModule}` | no path from user code — unreachable |
| mine path | `nil` | graph root — a legitimate entry point |

## Changes

`buildCallChains` now returns whether the chains came from an actual traceback; that fact already existed and was discarded.

`callGraphExportFinding` gains `Reachable *bool` — pointer because there are three states, and absent is a real answer, not a default. It is populated only when `ctx.userPackages != nil`, so the mine path emits no field rather than asserting `false` about a question that does not apply.

Entry-point construction skips findings whose `Reachable` is explicitly `false`, in both the streaming and non-streaming paths. A `nil` is exempt.

Schema bumped `6.6` → `6.7`; the field is additive and optional.

## Measured — `redis-demo`

| | before | after |
|---|---|---|
| crypto entry points | 7 | **6** (`Hashing$1.hash` removed) |
| TLS finding | no reachability field | `reachable: true`, 6-frame chain |
| MD5 finding | no reachability field | `reachable: false`, 1 frame |

Both findings are still reported. Unreachable crypto is marked, not omitted — it remains part of the dependency inventory, and some compliance regimes care that it ships at all. Only its promotion to entry point is withdrawn.

Mine-path regression check on `commons-codec 1.15` scanned standalone: **130 crypto entry points before and after**, and no `reachable` field on any finding.

## Tests

`TestBuildCryptoEntryPoints_ExcludesUnreachableFindings` covers all three states, including the mine-path exemption that makes the unconditional version of this change wrong.

`make lint` reports 0 issues. Full suite passes except the Postgres testcontainers group in `internal/engine`, which needs a running Docker daemon and fails identically on `main` in this environment.

## Notes

`docs/user-guide/user-guide.html` updated for the new schema version and field.

Two things deliberately left out. Entry points are still derived from one exported chain rather than from all reaching call sites, so `Jedis.set(...)` is absent although it reaches the same crypto as `Jedis.get(...)` — a separate defect in the tracer's frontier collapsing. And `findings.json` still carries no reachability information; this PR covers the callgraph export only.

## Summary by CodeRabbit

- **New Features**
  - Call-graph findings now indicate whether they are reachable from user code when applicable.
  - Call-graph export schema updated to version 6.7.
  - Expanded documentation for Java and Python cryptographic library support.

- **Bug Fixes**
  - Unreachable dependency findings are excluded from crypto entry-point results while library scans remain supported.
  - Improved Java cast handling, anonymous-class detection, resource/control-flow typing, and varargs fluent-chain resolution.

- **Documentation**
  - Documented the new conditional `reachable` field and its possible values.


agustingroh:
## Review follow-up

Confirmed and fixed in the latest commit — this was a real gap in the original change.

Supporting calls reach the entry-point index by two routes: through their finding graph, and through a fallback that sweeps up any the graphs left unclaimed. Skipping an unreachable finding removed the first route but left its supporting calls looking orphaned, so the fallback republished them as standalone entry points. The finding's containing function was withheld while its lifecycle came back through the side door.

Reproduced before fixing — the new subtest fails on the previous commit with:

```
entry points = [redis.clients.jedis.util.Hashing$1.hash(byte[]): long], want none
```

`redis-demo` did not expose it because its unreachable finding happens to have zero supporting calls, so the leak had no payload there. The subtest gives it one.

Fix is a `claimSupportingCalls` helper that marks the IDs as accounted for without indexing them, applied in both builders as suggested — the streaming path at the `else` of the reachability branch, and `buildCryptoEntryPoints` before the unreferenced sweep.

Re-verified after the change: `redis-demo` 6 entry points with `reachable` true/false as before, and the mine-path regression check on `commons-codec 1.15` standalone still reports 130 entry points against a 130 baseline. `make lint` 0 issues; `internal/scan`, `internal/callgraph` and `pkg/graphfrag` pass.

PR #250, opened 2026-08-13
fix: annotated declarations, expression receivers and lifecycle ordering

agustingroh:
## 1. Annotated declarations lost their visibility and start line

`Jedis.get` is declared `public` and was reported `package-private`, with `start_line` pointing at the `@Override` above it. Measured across jedis 5.1.0: **74.5% of methods declared `public` were not reported as public**, which makes any "is this a public entry point" reasoning unusable.

tree-sitter groups every modifier under one `modifiers` node, and the visibility scan compared that node's **whole text**:

```
no annotation:  modifiers = "public"           → matches
@Override:      modifiers = "@Override public" → matches nothing → package-private
```

It only ever worked for declarations carrying nothing else. The keyword is right there as a separate child (`modifiers → [marker_annotation] [public]`), so the scan now descends one level. `StartLine` skips leading modifiers for the same reason — the declaration line is what a reader jumps to, and what the bytecode index agrees with.

## 2. Expression receivers were adopted as type names

Every type lookup keys on a type **name**, so a receiver that is itself an expression missed all of them and landed in the final fallback, which took its source text as the class:

```
java.util.(key(key)).add#1          ×385
e.(getClass()).getSimpleName#0
```

**2693 such callees** in the graph fragment for `redis-demo` — the artifact the mining pipeline consumes. None names a class that exists anywhere.

An expression's result type is not resolved at this point, so the honest form is no type anchor at all. The call stays recorded, and the raw receiver text is preserved separately for fluent-chain resolution, but the identity no longer claims a class.

Measured on `redis-demo`: fabricated callees **2693 → 0**, `function_count` unchanged at 8266, and **internal edge count unchanged at 45130** — nothing was invented to replace them. Mine-path regression check on `commons-codec 1.15` standalone: 130 crypto entry points against a 130 baseline, 0 fabricated.

## 3. Supporting calls crossed a receiver reassignment

The TLS finding in `DefaultJedisSocketFactory.createSocket` collected `java.net.Socket.setSoTimeout` from line 90 — a plain-TCP call that runs **before** the `if (ssl)` block opens at line 92:

```java
socket = connectToFirstSuccessfulHost(_hostAndPort);
socket.setSoTimeout(socketTimeout);                                   // 90  plain socket
if (ssl) {
    _sslSocketFactory = (SSLSocketFactory) SSLSocketFactory.getDefault();  // 95  the finding
    socket = _sslSocketFactory.createSocket(socket, host, port, true);     // 97  socket rebound
    ((SSLSocket) socket).setSSLParameters(sslParameters);                  // 100
```

The lifecycle walk reached it legitimately — the finding assigns `_sslSocketFactory`, `createSocket` is a call on it, `createSocket` assigns `socket`, and `setSoTimeout` is a call on `socket` — but `selectDescendants` iterated calls with no order constraint, so a call from line 90 was attributed to an object that only exists from line 97 onward.

A name is not an identity. The walk now carries the index at which the variable was bound to the object being tracked and considers only calls after it. This is ordering within one function using an index that was already available, not flow analysis.

Result on `redis-demo`: supporting calls **4 → 3**, leaving exactly the three inside the `if (ssl)` block — `createSocket`, `setSSLParameters`, `getSession`.

## Combined effect — `redis-demo`

| | before | after |
|---|---|---|
| fabricated callees in fragment | 2693 | **0** |
| supporting calls on the TLS finding | 4 (one plain-TCP) | **3** |
| `Jedis.get` visibility / start line | `package-private` / 4922 | **`public` / 4923** |
| crypto entry points | 6 | 6 |
| findings, reachability, chain depth | 2, true/false, 6 | unchanged |
| graph edges | 45130 | 45130 |

## Tests

Three regression tests, each verified to fail without its fix (7 assertions across the two parser tests, 1 in the lifecycle test):

- `TestJavaParser_AnnotatedDeclarationsKeepVisibilityAndLine` — one, two and zero annotations, plus an annotated constructor
- `TestJavaParser_ExpressionReceiverIsNotATypeName` — call receiver and chained-call receiver
- `TestLifecycleCallIndices_ReassignedReceiverExcludesEarlierCalls` — the socket-rebinding shape

`make lint` reports 0 issues. Full suite passes except the Postgres testcontainers group, which needs a running Docker daemon and fails identically on `main` here.

One note for the record: `TestDependencyIntelligenceExportContract` failed once during a full-suite run and did not reproduce in three subsequent attempts (isolated, package-level, and two more full runs). It looks like a flake rather than a consequence of these changes, but it is worth watching in CI.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Java declaration visibility and source-line reporting, including annotated methods and constructors.
  * Corrected function identifier parsing for unanchored method names.
  * Prevented expression-based receivers from producing incorrect callee type names.
  * Fixed lifecycle call attribution after receiver variables are reassigned.
* **Tests**
  * Added regression coverage for declaration parsing, function identifier round-tripping, expression receivers, and reassigned receiver lifecycles.


agustingroh:
## Regression sweep — 6 libraries, 12 scans

Ran the mine path on six real Maven artifacts with `main` (4ec69ec) and this branch, comparing every metric. These are the artifacts the mining pipeline processes, so the sweep targets exactly what a re-mine would regenerate.

| library | assets | algo/proto names | entry points | supporting | functions | edges | fabricated | wall s |
|---|---|---|---|---|---|---|---|---|
| commons-codec 1.15 | 72→72 | 24→24 | 130→130 | 51→51 | 925→925 | 2758→2758 | 86→**0** | 19→19 |
| jedis 5.1.0 | 2→2 | 4→4 | 157→157 | 4→**3** | 8264→8264 | 18640→18640 | 2679→**0** | 67→67 |
| postgresql 42.7.3 | 43→43 | 8→8 | 212→212 | 44→44 | 3666→3666 | 13248→13248 | 530→**0** | 35→35 |
| kafka-clients 3.7.1 | 18→18 | 8→8 | 145→145 | 25→25 | 19687→19687 | 78211→78211 | 2534→**0** | 157→156 |
| google-auth-oauth2 1.19.0 | 8→8 | 11→11 | 18→18 | 8→8 | 1026→1026 | 2934→2934 | 496→**0** | 13→13 |
| guava 32.1.2-jre | 14→14 | 10→10 | 32→32 | 1→1 | 13103→13103 | 29599→29599 | 1976→**0** | 79→79 |

**No regressions.** Across 46,671 functions and 145,390 edges: no asset lost, no algorithm or protocol name lost, no entry point lost, `function_count` / `edge_count` / `internal_edge_count` byte-identical everywhere, and wall clock within 1 s.

**8,301 fabricated identities eliminated in total.** This was never a jedis quirk — every library carried them.

The only supporting-call movement is jedis 4→3, confirmed correct against source: the dropped call is `socket.setSoTimeout` at `DefaultJedisSocketFactory.java:90`, where `socket` still holds the plain socket bound at L89. It is rebound to the TLS-wrapped socket at L97, and the three retained calls are at L97/100/104 — all after the wrap.

Three callee names changed without any count changing, and in each case the previous name could not exist:

- postgresql `LibPQFactory.java:158` — `cf.toArray` → `toArray`; the receiver is `cf.generateCertificates(fis)`, a `Collection`, and `CertificateFactory` has no `toArray`
- kafka `DefaultSslEngineFactory.java:265` — `sslContext.getName` → `getName`; the receiver is `sslContext.getProvider()`, and `SSLContext` has no `getName`
- google-auth `TokenVerifier.java:387` — a malformed name spanning a line break collapses to `getPublicKey`

The visibility fix turns out broader than described: it also corrects declarations with several modifiers and no annotation (`Charsets.java:146`, `public static Charset toCharset(...)`, was reported `package-private`). Source agreement rose in every library — commons-codec 62%→93%, jedis 59%→73%, postgresql 76%→88%, kafka 81%→93%, google-auth 81%→93%, guava 79%→93% — and **zero** declarations went from correct on `main` to wrong here.

## Follow-up fix included

The sweep surfaced a defect introduced by the expression-receiver change, now fixed in this branch.

An unanchored callee has no package and no type, so `FunctionID.String()` renders it `.getSimpleName#0`. `ParseFunctionID` rejected that — its plain-function branch treats `lastDot == 0` as "no package separator" — so `recordEdgeResolution` could not recover the name and left `method_name` empty on every edge built from one. Measured on jedis: **2468 of 14454 external calls lost `method_name`**; after the fix, 0. `target_function_name` still carried the name, so nothing was unrecoverable, but a consumer reading `method_name` would have seen it vanish.

`String` and `Parse` are now symmetric for that form. Genuinely malformed input stays rejected, and `.pkg.name` is deliberately left alone — it already parsed as package `.pkg` before this change, and that pre-existing behavior is out of scope.

## Known behavior change, not fixed here

Because unanchored callees of the same method and arity now share a key, two such calls on one source line collapse into a single `external_calls` record. At `Rule.java:452`, `sb.append(", loc='").append(loc).append('\'')` yielded three records on `main` and yields two here. Sites affected: 9 / 638 / 86 / 148 / 43 / 52 across the six libraries.

**Zero of them had a legitimate type anchor on `main`** — every one was a fabricated identity — and zero sites were gained, so no real call site was lost. The net trade is 8,301 fabricated identities removed against ~976 merged records among calls that never resolved. Worth stating explicitly since this feeds the mining pipeline.

PR #310, opened 2026-08-26
feat(python): callgraph parser parity with Java (binding resolution, synthetic entry points, imports)

matiasdaloia:
## Summary

Python callgraph parser parity with Java — the first change in the per-language parity series (Go, Rust, Node follow using these two SDD changes as the template). Serves both consumption paths: local scans (with and without `--scan-dependencies`) and mining via crypto-mining-service. No export schema change (`6.13` / `graph-fragment-1.13`); `internal/scan/supporting_calls.go` untouched.

### Phase 1 — `python-parser-java-parity` (archived)
| Row | Capability |
|---|---|
| 1 | Function parameters usable as call receivers |
| 3 | `with/for/except … as`, async variants, walrus, tuple/star unpacking bind receivers |
| 2 | `self.attr` / `cls.attr` provenance across a class's methods |
| 12 | Synthetic `<module>` / `<clinit>` entry points for module-level and class-body crypto calls |
| 4 | Imports inside `try/except`, `if TYPE_CHECKING`, function bodies; relative imports resolved against the package |
| 15 | `__init__.py` re-exports stitched to the declaring module (project-local only — KB-keyed dependency paths never rewritten) |

### Phase 2 — `python-parser-parity-2`
| Area | Capability |
|---|---|
| Perf | Single-descent parser with deferred per-scope call resolution; `NamedChild` traversal; symbol-id dispatch. **0.98× the pre-parity baseline** (was ~+100% after phase 1 draft). CI guard `TestPythonParser_NodeVisitBudget` |
| 6 | Opengrep column-convention pinning test for Python (runs in CI via `Dockerfile.test`) |
| 7 | Bounded dynamic dispatch: `getattr(obj, "lit")(…)`, `importlib.import_module("lit")`, `__import__("lit")` — literal strings only |
| 8 | Decorator-aware receivers (`@staticmethod`, `@classmethod` `cls`, `@property`, custom) — also fixed a pre-existing `cls.foo()` mis-typing |
| 9 | `super().__init__()` / `super().m()` resolve to the first base |
| 11 | `functools.partial(f, …)` and implicit `__call__` resolve to their targets; `__call__`/`__enter__`/`__exit__` now declared |
| 13 | Type hints feed receiver resolution: param/return annotations, `Optional`/`Union`/`X | None`/forward refs/`typing.` prefix, annotated assignment, TYPE_CHECKING imports |
| 14 | `PythonDependencyTypeResolver`: signature index from `.pyi` / annotated `.py` of pip-resolved distributions, disk-cached, KB-first fallback chain |
| 18 | Visibility / owner visibility from naming convention |
| 20 | Nested constructor-argument provenance (`ArgumentSources`) |
| KDF | `resolved_key_length` populated for Python KDFs (existing 6.13 field): pyca-cryptography, hashlib (new KB), argon2-cffi, bcrypt, pycryptodome/x — keyword and positional forms; signatures verified against upstream source (fixed pre-existing arity bugs: `Scrypt` 4→5, pycryptodome `scrypt` 4→6, `HKDF` 3→4) |

### Review
Two fresh-context reviews (phase 1: F1–F9; phase 2: G1–G13) — all findings fixed test-first; see the comment thread.

### Verification (HEAD)
- `go test -race ./...` ok · `make lint` 0 issues · `make coverage-check` 82.2% (≥80%) · `git diff --check` clean
- Zero diff under `pkg/graphfrag/`, `internal/scan/supporting_calls.go`
- Benchmark `BenchmarkPythonParseDirectory_Bindings`, 8 reps: 66.7 ms/op (0.98× `c6ee180`), 20.8 MB/op (1.08×)

### SDD artifacts
`openspec/changes/archive/2026-08-27-python-parser-java-parity/`, `openspec/changes/python-parser-parity-2/` (archived on merge-readiness), main specs under `openspec/specs/`.

Size: `size:exception` accepted at SDD preflight (single PR, no split).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


matiasdaloia:
## Fresh-context review — findings being fixed on this branch

Gates at review time: `go test -race` ok, `make lint` 0 issues. Benchmark (200-module corpus): merge-base ≈73 ms/op, 314k allocs/op → HEAD ≈132 ms/op, 615k allocs/op (**+80% time, +96% allocs** vs 10% design ceiling).

| # | Sev | Finding | Fix |
|---|---|---|---|
| F1 | HIGH | `rewritePythonReExportedCallee` rewrites KB-keyed public paths for sub-package re-exports (e.g. `authlib.jose.JsonWebSignature` → `authlib.jose.rfc7515.…`), breaking contract lookup when dependency sources are in the graph | Restrict re-export stitching to project-local analyses; negative test with the authlib shape |
| F2 | HIGH | Each method body traversed ~6× (unpruned module/class binding walks, separate class-attr walk, recursive imports, per-method bindings + calls); cgo `Child`/`Type`/`ChildCount` dominate | Two-pass design: one scope-stacked pre-pass (imports, re-exports, class attrs, per-scope binders) + one call pass; symbol-id compares instead of `Type()` strings. Target ≤10% over merge-base |
| F3 | MED | `<module>`/`<clinit>` decls take `ReceiverVar` from names bound only inside nested functions | Scope stack in F2; negative test |
| F4 | MED | UPPER_CASE locals never become receivers (CapitalCase heuristic runs before locals check) | Locals before heuristic; test |
| F5 | MED | Spec describes flat `pkg/mod.py` re-export; test uses `pkg/mod/__init__.py` (flat case is already resolved, no rewrite) | Align spec; add flat-layout test |
| F6 | LOW | `import a.b.c` binds `Imports["a"]="a.b.c"` → `a.b.c.foo()` resolves to `a.b.c.b.c.foo` | Bind `a`→`a` |
| F7 | LOW | Empty `packagePath` → `.stem`; `.pyi` stem collides with `__init__.py` `<module>` | Guard + distinct stem |
| F8 | LOW | Duplicate `### Fixed` heading; single 1,100-char paragraph; user guide not updated | Merge/split bullets; guide note |
| F9 | LOW | Missing tests: param shadowing import, `*args/**kwargs`/defaults, `self.a.b`, `self` in free function | Add |

Checked clean: `functionTypeClassInit` rename call sites, `isFunctionContainer` Python-only node names, `builder.go` `addAnalyses` split, no `internal/failure` import, commit hygiene.

matiasdaloia:
## Review findings addressed

All F1–F9 fixed on this branch (`5d49e19..6d7ba35`), test-first.

| # | Outcome | Commit |
|---|---|---|
| F1 re-export rewrite vs KB paths | Re-export stitching gated to project-local analyses; negative test `TestBuilder_InitPyReexport_DoesNotRewriteKBKeyedDependency` (authlib shape) | `5d49e19` |
| F2 parse overhead | Single-traversal prepass (`collectPythonFilePrepass`) for imports, re-exports, class attrs, per-scope locals; `Node.Symbol()` instead of `Node.Type()`; cached child counts | `c4e003e`, `b7b39e9` |
| F3 nested-scope leak into `<module>`/`<clinit>` | Fixed by the scope-stacked prepass; `TestPythonParser_SyntheticEntryPoint_NoNestedScopeLeak` | `c4e003e` |
| F4 UPPER_CASE locals | Locals checked before the CapitalCase heuristic | `94c16d7` |
| F5 spec/test drift | Spec covers flat (already resolved) and sub-package (rewrite) layouts; `TestBuilder_InitPyReexport_FlatLayoutAlreadyResolved` | `c8a0e93` |
| F6 `import a.b.c` binding | Binds the top-level name only | `94c16d7` |
| F7 module dotted path | Empty package path and `.pyi` stems handled | `5a947eb` |
| F8 CHANGELOG / user guide | Single `### Fixed` section, one bullet per row; user-guide reachability note | `825d733` |
| F9 test gaps | Param shadowing import, `*args/**kwargs`/defaults, `self.a.b`, `self` in free function | `494302c` |
| bonus | Pre-existing non-deterministic `goconst` lint hit removed | `a1dca1d` |

**Benchmark** (`BenchmarkPythonParseDirectory_Bindings`, 200-module corpus, 8×2s per side, baseline in a fresh worktree at `c6ee180`):

| | ns/op | B/op | allocs/op |
|---|---|---|---|
| c6ee180 | 73.15 ms | 19.27 MB | 313,968 |
| HEAD | 83.39 ms | 25.99 MB | 284,670 |
| Δ | **+14.0%** | +34.9% | −9.3% |

Within the 20% hard cap; above the 10% design target. Remaining cost is the per-function call walk plus KB inference over the ~25% additional declarations the feature legitimately extracts. Machine noise across rounds was material (14–36%), so treat the 8-rep figure as the reference.

Gates: `go test -race ./...` ok · `make lint` 0 issues · coverage 82.0% · `git diff --check` clean · zero diff under `pkg/graphfrag/`, `internal/scan/supporting_calls.go`, `internal/callgraph/contracts/`.

matiasdaloia:
## Phase 2 in progress — `python-parser-parity-2` (same branch)

Scope: close the perf gap for mining throughput and land every deferred parity row so Python reaches Java's level. SDD artifacts under `openspec/changes/python-parser-parity-2/`.

**Perf gate met** (`BenchmarkPythonParseDirectory_Bindings`, 8 reps, same session, baseline in a fresh worktree at `c6ee180`):

| Stage | ns/op | B/op |
|---|---|---|
| `c6ee180` baseline | 67.9 ms (1.00×) | 19.28 MB (1.00×) |
| phase-1 head (before) | ~14–20% over | ~1.35× |
| single-descent parser (`32d9791`) | 60.6 ms (0.89×) | 20.39 MB (1.06×) |
| after rows 6/18/20/8/9 (`e49e921`) | 67.3 ms (0.99×) | 20.57 MB (1.07×) |

Python parsing is now at or below the pre-parity baseline while extracting ~25% more declarations. CI guard: `TestPythonParser_NodeVisitBudget` fails on traversal regressions.

Landed so far: row 6 (opengrep column-convention pinning for Python), 18 (visibility from naming convention), 20 (nested constructor-argument provenance), 8 (decorator-aware receivers — also fixed a pre-existing `cls.foo()` mis-typing), 9 (`super()` resolves to the first base).

In progress: 7 (bounded dynamic dispatch), 11 (`functools.partial`/`__call__`), 13 (type hints), KDF `resolved_key_length` for Python (existing 6.13 field, no bump), 14 (dependency-mode type resolver from `.pyi`/annotated sources). Final gates, CHANGELOG, and user guide follow.

matiasdaloia:
## Fresh-context review — phase 2 (`73c9d28..29ceef5`) — findings being fixed on this branch

Gates at review: `go test -race` ok; `make lint` 2 issues (1 introduced here); benchmark 1.00× ns/op, 1.08× B/op vs `c6ee180`.

| # | Sev | Finding | Fix |
|---|---|---|---|
| G1 | HIGH | `propagatePythonAssignedVarTypesForDecl` rewrites `Callee.Package` to the caller's package and `Type` to the raw `ReturnType` → FQNs that exist nowhere (`app.(Cipher).encrypt`), and KB `<init>`→`encryptor` chains stop matching in dependency scans | Split dotted type into `Package`/`Type`; otherwise take `Package` from the declaring decl; builder-level test with a versioned dep |
| G2 | HIGH | KDF step 3b treats a keyword arg sitting at the keySize index as a positional constant (`PBKDF2(pw, salt, count=1000)` → 8000 bits) | Skip keyword-shaped expressions in 3b; negative test |
| G3 | HIGH | `parseFunctionDef` drops all dunders except `__init__`, so `Class.__call__` (row 11) has no declaration → dangling edge | Whitelist `__call__` (+ `__enter__`/`__exit__`); reachability test through `__call__` |
| G4 | MED | Inner `super()` still emitted as `mypkg.super` (spec: never) | Suppress the inner node; extend test |
| G5 | MED | Dynamic imports registered at resolution time — invisible to already-resolved function scopes; `import_module(".mod", "pkg")` yields a junk key | Register at walk time; skip relative literals |
| G6 | MED | Module-constant provenance ignores local shadowing / later rebinding → fabricated `resolved_key_length` | Respect local layer; drop on non-integer rebinding |
| G7 | MED | Step 3b is additive for non-Python ecosystems; comment/test claim otherwise | Gate on ecosystem or document + pin |
| G8 | MED | `make lint` red: `prealloc` introduced by `9aa415c`; pre-existing `goconst` in `builder.go` | Fix both |
| G9 | LOW | Symbol resolution lacks `SymbolTypeRegular` filter (`type`/`await`/`lambda` collide) | Filter + assert |
| G10 | LOW | Row 13: `typing.Optional[...]` not normalized; annotated type survives un-annotated rebinding | Normalize `typing.` prefix; clear on rebinding |
| G11 | LOW | CHANGELOG has duplicate `### Fixed`/`### Added` headings; perf bullet wording | Merge; reword |
| G12 | LOW | User guide still cites schema `6.12`/`graph-fragment-1.12` | Update to 6.13 / 1.13 |
| G13 | LOW | Stale doc comment referencing removed `collectPythonFilePrepass`; opengrep absent in CI so the column pin skips there; signature-cache errors lack `callgraph:` prefix; cache dir created without `--scan-dependencies` | Fix comment/prefix/lazy dir; add opengrep to `test.yml` |

Checked clean: `decl.Calls` document order, scope stack, decorators, visibility, `ArgumentSources` depth, byte→bit overflow guard, every YAML `parameter_types` length == arity with indices matching verified signatures, `NodeVisitBudget` non-skipping.

matiasdaloia:
## Phase-2 review findings addressed

All G1–G13 fixed test-first, one commit each (`ac06f87..4f4f72c`).

| # | Outcome | Commit |
|---|---|---|
| G1 type propagation FQNs | Dotted return types split into Package/Type; bare names take the declaring decl's package; builder + KB-chain tests | `ac06f87` |
| G2 keyword arg at keySize index | Step 3b skips keyword-shaped expressions; negative test | `efeb4b8` |
| G3 `__call__` decl dropped | `__call__`/`__enter__`/`__exit__` whitelisted; reachability test through `__call__` | `be51004` |
| G4 inner `super()`/`getattr` node | Suppressed when composing an enclosing expression | `86e2fda` |
| G5 dynamic import timing | Registered at walk time; relative literals skipped | `9f2aeaa` |
| G6 module-const shadowing/rebinding | Local layer respected; non-integer rebinding drops the entry | `4137f5c` |
| G7 step 3b for non-Python | Gated to the Python ecosystem; Java test extended | `efeb4b8` |
| G8 lint | `make lint` 0 issues (also consolidated a pre-existing `goconst`) | `6bd7727` |
| G9 symbol table | `SymbolTypeRegular` filter, unit-tested | `2cf0f7d` |
| G10 `typing.` prefix / rebinding | `typing.Optional/Union` (a `subscript` node) normalized; un-annotated reassignment clears the type | `f577317` |
| G11/G12 docs | Single `### Added`/`### Fixed`; user guide schema numbers → 6.13 / 1.13 | `f81c580` |
| G13 misc | Stale comment fixed; signature cache lazy dir + `callgraph:` error prefix. CI-opengrep item was a false premise — `Dockerfile.test` already installs opengrep | `781e8fe`, `be51004` |

**Perf** (8 reps, one run): 66.7 ms/op (0.98× `c6ee180`), 20.8 MB/op (1.08×). Budget ≤1.10× / ≤1.15×.
**Gates**: `go test -race ./...` ok · `make lint` 0 issues · coverage 82.2% · `git diff --check` clean · zero diff under `pkg/graphfrag/`, `internal/scan/supporting_calls.go` · schema 6.13 / graph-fragment-1.13 unchanged.

PR #332, opened 2026-08-31
fix(callgraph): skip condensed routes above path-count ceiling

### LINKED ISSUE #292: Condensed call-chain export builds the full path set before applying maxChains, exhausting memory
## Problem

The condensed call-chain export applies its `maxChains` budget to what it
*emits*, not to what it *builds*. `TraceBackCondensed` materialises the complete
reachability structure and counts every path before truncating the result:

```go
// internal/callgraph/tracer.go
reach := graphwalk.Reach(targetKey, t.walkOptions(userPackages, maxDepth))
condensed := graphwalk.Condense(reach, nodeLess)

total = graphwalk.Count(reach, condensed)                        // full path count
for _, route := range graphwalk.Routes(reach, condensed, maxChains) {
    chains = append(chains, t.materializeRoute(route))            // capped here
}
return chains, total, len(chains) < total
```

For a densely-connected function the path count is combinatorial in the graph's
fan-in and depth. The budget is spent building a structure that is then thrown
away, so peak memory is governed by the graph's true path count rather than by
the configured limit — and the limit gives no protection at all against the case
it exists for.

## Why this matters

The failure mode is process death, not a degraded result. The scan is killed by
the OS before it can report anything, so:

- The whole component fails, including the findings that had already been
  computed. One pathological function costs the entire scan.
- The caller sees `signal: killed` with no actionable detail. It is not
  distinguishable from an unrelated crash, and it is not retryable — the same
  input reproduces it.
- `structuralTracebackChains` memoises per function and the export iterates every
  function carrying a finding, so the cost accumulates across a single export
  rather than peaking once.

This lands hardest on large, heavily-reflective libraries — exactly the
components where reachability analysis is most valuable, because a human cannot
trace those paths by hand.

## Evidence

Observed during a large-scale batch scan, from a component that was killed mid-export:

```
inference: stats origin_counts=map[constructor:32 propagated:6] join_failed=5 suppressed=9189
WRN Truncated condensed call chain export for finding function
    emitted=128
    function=org.codehaus.groovy.runtime.(EncodingGroovyMethods).digest#2$byte[],String
    max_chains=128
    total_condensed_paths=473526
: signal: killed
```

473,526 paths were counted and structured to emit 128 — a factor of ~3,700 of
wasted work, for one function. Reproduced across three versions:

| component |
|---|
| `pkg:maven/org.codehaus.groovy/groovy-all@3.0.25` |
| `pkg:maven/org.apache.groovy/groovy@4.0.22` |
| `pkg:maven/org.apache.groovy/groovy@5.0.3` |

Note the warning is emitted *before* the process dies, so the truncation path is
working as designed. The defect is that reaching the point of being able to
truncate is what exhausts memory.

## Options

**1. Defensive cap before expansion (smallest change).** If `Count` exceeds a
threshold, skip `Routes` entirely: emit zero chains and set `truncated`. The
existing contract already expresses this — `analysis.call_chains` goes partial
and reachability is downgraded to `unknown` — so nothing downstream needs to
change. This trades a small amount of detail on pathological functions for not
dying, and it unblocks the affected components immediately.

**2. Bound the walk itself (correct fix).** Thread the chain budget into
`Reach` / `Condense` so the structure is never larger than what will be
consumed. This is the real fix and it removes the waste rather than detecting it
after the fact, but it changes `graphwalk`'s contract and needs its own
correctness work — a bounded walk must still pick a *representative* set of
paths, not merely the first ones it happens to find.

**3. Per-job memory ceiling (defence in depth, integrating systems).** A resource cap
in the integrating worker so a pathological component fails as a clean, categorised
error instead of being killed by the OS. Worth doing regardless of 1 and 2,
because it converts an unexplained crash into a diagnosable one.

Suggested sequencing: land 1 to stop the bleeding, track 2 as the real fix, and
raise 3 separately against the worker.

## Related

Components that exhaust memory here overlap with those large enough to hit the
graph-size storage ceiling tracked in downstream storage-safety work. The two are
independent defects with a shared trigger: unusually large call graphs. Fixing
either one alone still leaves those components unanalysable.

matiasdaloia:
## Summary
- Closes #292 by applying the defensive path-count ceiling (option 1): when condensed `Count` exceeds 100000 for a finding, skip `Routes`, emit no chains, and keep `truncated`/partial so pathological fan-in cannot exhaust memory after counting a huge route set.
- Live `--export-callgraph` and stitched export share `graphwalk.PathCountSkipThreshold`.
- Documents the existing per-finding bounds (128 chains, 32 frames) and the new 100000-route skip ceiling in the user guide and changelog.

## Test plan
- [x] `go test ./internal/callgraph/ -run TestTraceBackCondensed_ -count=1`
- [x] `go test ./pkg/graphwalk/ ./pkg/graphfrag/ ./internal/scan/ -count=1`
- [x] User-guide HTML parse + prohibited-term grep + `git diff --check`
- [ ] CI green on this PR


matiasdaloia:
Follow-up commit addresses blind self-review blockers:

- Live export no longer stamps `reachable: false` or synthesizes a one-node self-chain when the path-count ceiling skips Routes; `reachability` stays `unknown`, `analysis.call_chains` stays `partial` (including mine path), and `call_chains` stays empty.
- Stitch emits a `PathCountTruncated` finding carrier instead of a fabricated self-entry; `ToCallgraphExport` publishes empty chains with unknown/partial.

Regression coverage added for both consumers.

PR #370, opened 2026-09-01
Fix/rust semantic identity gaps

agustingroh:


agustingroh:
## What `trait_associated_types` is, with an example

This is a new schema field, not something that existed before this PR. Rust lets a trait declare an *associated type* — a type each implementer must name, distinct from a method:

```rust
trait KeyInit {
    type KeySize;
    fn new(key: &GenericArray<u8, Self::KeySize>) -> Self;
}
```

Generic code written against the trait references that associated type without knowing which concrete algorithm fills it in:

```rust
fn setup<C: KeyInit>(bytes: &[u8]) -> C::KeySize { ... }
```

Before this PR, the KB only knew how to describe **methods** (`method + arity -> return type`). There was no way to say "trait X declares an associated type named Y", so `C::KeySize` resolved to nothing. `trait_associated_types` adds exactly that:

```yaml
trait_associated_types:
  - trait: KeyInit
    name: KeySize
    type: generic_array::GenericArray
```

Notes on scope, since this touches the shared contracts loader (`internal/callgraph/contracts/contracts.go`):
- The field is `omitempty` — every existing Java/Go/Python/C/C++ YAML is unaffected; they simply don't declare it.
- `KnowledgeBase.TraitAssociatedType()` is called only from `rust_type_semantics.go` today — no other language's parser reads this data.
- `LoadEmbedded(ecosystem)` loads and merges YAMLs per ecosystem directory, so a Rust entry can never collide with or leak into another language's KB.

`contracts: []` is intentionally empty in the four new YAMLs (`crypto-common`, `digest`, `aead`, `elliptic-curve`): those crates declare the base traits (`KeyInit`, `Digest`, `AeadCore`, `Curve`) that the real algorithm crates (`aes`, `sha2`, `chacha20poly1305`, `p256`, ...) implement. They carry no terminal crypto operation of their own — only the associated-type facts above.

## A real-world case: a generic hash function

This is an idiomatic RustCrypto pattern, not a contrived example — writing one hash function that works with any algorithm implementing `Digest` (`Sha256`, `Sha512`, `Blake2`, ...) instead of one per algorithm:

```rust
use digest::Digest;
use generic_array::GenericArray;

fn hash_data<D: Digest>(data: &[u8]) -> GenericArray<u8, D::OutputSize> {
    let mut hasher = D::new();
    hasher.update(data);
    hasher.finalize()
}

fn main() {
    let digest = hash_data::<sha2::Sha256>(b"hello world");
    println!("{:x}", digest);
}
```

`D::new()`, `hasher.update(data)` and `hasher.finalize()` already resolved correctly before this PR — `D` was already identified as `Digest` via its trait bound. What didn't resolve was `D::OutputSize` in the return type: it carried the trait's own name where a real type belongs, instead of a real identity. That matters because the output/key size is exactly the kind of data crypto-finder's contracts report (`resolved_key_length`) — a generic function reporting or forwarding that size carried a low-quality identity in the graph instead of a real one.

The closest real, *measured* instance of this shape is `elliptic_curve::Curve::FieldBytesSize` (used generically across `p256`, `p384`, `k256`, `ecdsa`):

```rust
use elliptic_curve::Curve;

fn element_size<C: Curve>() -> C::FieldBytesSize { /* ... */ }
```

This one isn't hypothetical — it's the exact shape recorded in this repo's own historical measurement as a declined gap (`C::FieldBytesSize::method`, 47 edges across a 53-published-crate corpus, `openspec/specs/rust-callgraph-identity-resolution/spec.md`, "Decisions deliberately not implemented"). `trait_associated_types` is what let this PR reopen and close that specific gap with real evidence, rather than reversing the earlier decision on a hunch.

PR #389, opened 2026-09-01
fix(scanner): resolve ambiguous file extensions by reading content

agustingroh:
## The bug

Several file extensions map to more than one language. Asked by **filename alone**, `enry` breaks the tie without looking at the file, and three of its answers are wrong:

| file | candidates enry sees | answered | correct |
|---|---|---|---|
| `.rs` | RenderScript, Rust, XML | **`xml`** | `rust` |
| `.cs` | C#, Smalltalk | **`smalltalk`** | `c#` |
| `.php` | Hack, PHP | **`hack`** | `php` |

`.go`, `.py`, `.java`, `.rb`, `.kt`, `.scala`, `.swift` and the rest are unaffected — they are unambiguous.

The wrong name is stored on **every finding** for those languages and travels downstream into persisted results.

## Why nobody noticed

It fails quietly in both directions. `xml` is not an ecosystem any consumer here recognises, so `ecosystemFromHint` returns `""` for it and the finding contributes *nothing* rather than erroring. And in practice the occurrence-key ecosystem is supplied by the explicit `--languages` flag appended afterwards, so the fallback covers for it. Nothing ever surfaced the mistake.

## The fix

`detectLanguage` now passes the file's leading bytes to enry, which resolves all three. The read is capped at 16 KiB — enry's content strategies look at shebangs, keywords and early declarations, all near the head of the file — so a large generated source costs a bounded read.

An unreadable path still falls back to the name-only answer: a possibly-wrong language is more useful downstream than none, and it is what keeps the existing name-only table test meaningful (it passes paths that do not exist).

## The one place this changes an answer something already relied on

Reading content also lets a **C++ header be told apart from a C one** — a `.h` with C++ content now reports `c++` where it used to report `c`.

`reportOccurrenceKeyEcosystems` deliberately added **both** the `c` and `cpp` ecosystems for anything detected as `c`, with a comment saying why: *"a C-language detector can report both C and C++ sources … so a C++ header never hides `.c` anchors."* A `.h` reporting `c++` would have fallen through to the generic branch and added **only** `cpp` — dropping the very ecosystem that branch exists to guarantee.

So the branch now tests the **C family** rather than the literal string. `TestReportOccurrenceKeyEcosystems_CPPHeaderKeepsBothEcosystems` is a characterisation test written to pass *before* the detector changed, and it fails on `c++` against the old branch — so this pairing cannot be silently undone.

## Verification

- Unit tests for all three ambiguous extensions, both `.h` directions, the unreadable-path fallback, and a file deliberately larger than the sniff cap.
- End-to-end: scanning a real Rust crate now emits `"language": "rust"` where it previously emitted `"language": "xml"`.
- Full suite unchanged from `origin/main`: **31 packages ok, 0 failures**. `go vet` and `gofmt` clean.

## Note on existing data

This fixes the detector, not rows already written. Components persisted before this change keep the old value until they are next scanned. That is left deliberately — the stored value corrects itself when those components are re-scanned for other reasons.

### DIFF
--- a/internal/cli/scan.go
+++ b/internal/cli/scan.go
@@ -596,8 +596,8 @@ func reportOccurrenceKeyEcosystems(target string, report *entities.InterimReport
 
 // isCFamilyLanguage reports whether a detected language is C or C++, which
 // share headers and translation units and so share the anchor treatment above.
-func isCFamilyLanguage(language string) bool {
-	switch language {
+func isCFamilyLanguage(lang string) bool {
+	switch lang {
 	case "c", "c++", ecosystemCPP:
 		return true
 	default:
--- a/internal/cli/scan.go
+++ b/internal/cli/scan.go
@@ -568,9 +568,15 @@ func reportOccurrenceKeyEcosystems(target string, report *entities.InterimReport
 	}
 	add := func(hint string) { addEcosystem(ecosystemFromHint(target, hint)) }
 	for _, finding := range report.Findings {
-		if finding.Language == "c" {
+		if isCFamilyLanguage(finding.Language) {
 			// A C-language detector can report both C and C++ sources. Build both
 			// source-only parsers so a C++ header never hides .c anchors.
+			//
+			// This tests the C FAMILY, not the literal "c". Once the detector
+			// reads file content it can tell a C++ header from a C one and
+			// reports "c++" for the former, which used to fall through to the
+			// generic branch below and add only cpp -- dropping the c ecosystem
+			// this branch exists to guarantee.
 			addEcosystem("c")
 			if isCPPSource(finding.FilePath) || cppHeaderTarget(target) {
 				addEcosystem(ecosystemCPP)
@@ -586,6 +592,17 @@ func reportOccurrenceKeyEcosystems(target string, report *entities.InterimReport
 		add(ecosystemFromHints(target, languageHints))
 	}
 	return ecosystems
+}
+
+// isCFamilyLanguage reports whether a detected language is C or C++, which
+// share headers and translation units and so share the anchor treatment above.
+func isCFamilyLanguage(language string) bool {
+	switch language {
+	case "c", "c++", ecosystemCPP:
+		return true
+	default:
+		return false
+	}
 }
 
 func occurrenceAnchorKey(ecosystem, functionID string) string {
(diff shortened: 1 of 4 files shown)


agustingroh:
## Differential over 85,662 real files, in answer to "does this break other languages?"

Rather than reason about hand-picked cases, I ran the old and new detector side by side over every file in 200 extracted crates, the four SCANOSS repos and the whole Go module cache.

**85,662 files scanned. 7,170 change (8.4%), across 31 distinct transitions.**

What matters is not how many change, but how many cross the set of languages `ecosystemFromHint` actually recognises (`c`, `c++`/`cpp`, `go`, `java`, `python`, `rust`, `node`/`javascript`/`typescript`). Everything else resolves to `""` today and after, so it has no behavioural effect.

| bucket | files | detail |
|---|---|---|
| **Gain** a supported ecosystem | 5,022 | `xml → rust` — the bug this PR is about |
| **Move within** the supported set | 207 | `c → c++`, all `.h` files. Covered by the `isCFamilyLanguage` change, so both `c` and `cpp` are still added |
| **Lose** a supported ecosystem | **0** | — |
| No effect (unsupported → unsupported) | 1,936 | e.g. `motorola 68k assembly → unix assembly`, `json → oasv2-json`, `perl → raku` |

### The differential found a real regression, which is now fixed

The first version of this PR had **one** file losing a supported ecosystem: a generated tree-sitter `parser.c` in the module cache went from `c` to `unknown`. Cause: `enry.IsBinary` returns true for its dense state tables, and `GetLanguage` answers `""` for content it thinks is binary.

Fixed in `70ea1a7`: each lookup is retried without content whenever the content-informed one comes back empty, so **content can only refine an answer, never erase one**. `TestDetectLanguageDoesNotLoseAnAnswerToBinaryHeuristics` pins it.

That also collapsed five other transitions and took the loss column to zero.

### The five `.inc` files, for completeness

Before the `70ea1a7` fix there were also five `.inc` files moving out of `c++` (to `assembly`, `pov-ray sdl`, `php`). Those are gone too, but they were never reachable anyway: `detectLanguage` only runs on files that produced semgrep results, and scanning those exact files with `--languages cpp` yields **0 findings**, because `.inc` is not a scanned C/C++ extension.

### Also

`make lint` with the repo's pinned golangci-lint: **0 issues**. Full suite: **31 packages ok, 0 failures**, unchanged from `origin/main`.

The differential harness was a throwaway and is not part of the diff.

PR #396, opened 2026-09-01
fix(rust): stop a crate's rules claiming the consumer's own receiver types

agustingroh:
## The problem

A Rust rule identifies its crate with a file-level import guard:

```yaml
- pattern-inside: |
    use $IMPORT;
    ...
- metavariable-regex: {metavariable: $IMPORT, regex: '(?i)^(ed25519.dalek).*'}
- pattern-either:
    - pattern: $KEY.sign($MSG)
```

The guard anchors the **file**. It does not anchor the **call**: `$KEY` is a free
metavariable, so in any file that imports the crate the rule also claims a
same-named method on a type the consumer declares itself.

```rust
use ed25519_dalek::SigningKey;        // an ordinary, genuine import

struct ConsumerOwnType;
impl ConsumerOwnType { fn sign(&self, m: &[u8]) -> Vec<u8> { m.to_vec() } }

let own = ConsumerOwnType;
let _ = own.sign(b"x");               // reported as an ed25519-dalek signature
```

That attribution reaches the served report as a fact about the component. The
construct is used by 60 of the 70 Rust crates in the ruleset; 28 of those have at
least one rule matching a method on a metavariable receiver.

The rule files have recorded the limit since they were written — p256's says a
receiver's type "is not knowable from the syntax tree", ed25519-dalek's adds that
constraining it further "is not possible at this layer". Both are true of the
rule layer. Neither is true here.

## What this does

Two commits.

**1. `fix(callgraph)`** — a pre-existing defect that had to go first.
`resolveFluentChainsByReturnType` re-qualifies each link of a method chain from
the previous link's return type, through `normalizeLookupTypeName`, which reduces
`ed25519_dalek::Keypair` to the bare `Keypair`; it then looks that bare name up in
an index built **only** from types the scanned source declares. A wrapper crate
that names its own type after the one it wraps therefore had a correctly-resolved
dependency callee overwritten with its own same-named method. On tlfs-crdt 0.1.0
`src/crypto.rs:39` (published on crates.io):

```rust
impl Keypair {                                    // the wrapper's own type
    fn to_keypair(self) -> ed25519_dalek::Keypair { .. }
    pub fn sign(self, msg: &[u8]) -> Signature {
        self.to_keypair().sign(msg)               // <- this link
    }
}
```

The parser resolves that link to `ed25519_dalek.(Keypair).sign`; the pass rewrote
it to `tlfs_crdt::crypto.(Keypair).sign`. The call then joins the wrapper's
contract instead of the crate's. The pass now carries the qualifier the return
type was *declared* with and declines to re-qualify when it names a different
unit. An unqualified return type keeps the previous behaviour exactly, so the
Java fluent chains this pass was written for are untouched.

**2. `feat(scan)`** — `FilterForeignReceiverAssets` drops an asset only when all
four hold; anything unresolved keeps it, as `deadcode.FilterReport` keeps
findings in a file it cannot read:

1. the rule ID names a crate;
2. the match resolves to at least one call node;
3. no node at that position resolves into the claimed crate, and one resolves to
   a type **declared in the scanned source**;
4. that declaring crate differs from the crate the rule names.

Condition 3 judges **every** node at the match position rather than the one
`findCryptoCallNode` returns, because that one is the chain ROOT by contract and
in a chain the root is frequently the consumer's own type while the crypto call
is a later link. Condition 3 is also what makes scanning the library itself safe:
inside ed25519-dalek, `impl SigningKey { fn sign }` is source-declared under
`ed25519_dalek`, so condition 4 fails.

Scope is Rust, deliberately: the import-guard construct is 100% Rust (741
occurrences of `use $IMPORT;`, no other language uses it), and every other
ecosystem is already mined and served.

## Why not fix it in the rules

Converting the guarded forms to `mode: taint` sourced at the crate's
constructors was measured, on the pinned opengrep (1.12.1, what production mines
with), over published reverse-dependency corpora. It removes the false matches
but only reaches receivers a same-file constructor flows into, so it loses real
detections wherever the key arrives from a struct field, a parameter or a
function return:

| crate | false matches | real sites lost |
|---|---|---|
| ed25519-dalek | 6 → 0 | 8 |
| tokio-rustls | 3 → 1 | 9 |
| p256 | 13 → 0 | 6 |
| russh | 9 → 0 | 6 |
| x25519-dalek | 1 → 0 | 2 |

The lost sites are true positives in wrapper crates. Filtering here costs no
recall, and additionally keeps the field receiver taint drops.

## Measurements

Everything below runs `crypto-finder scan --no-remote-rules --rules-dir <ruleset>`
against the WHOLE Rust ruleset — all 618 rules at once, not one crate's rules
against its own corpus — which is the invocation the mining service uses.

### Which rules can be affected at all

| | |
|---|---|
| Rust rules | 618 |
| with a free-receiver pattern (`$X.m(..)` or `$T::m(..)`) | 190 |
| — of those, `mode: taint` | **128, structurally immune** |
| — of those, anchored by an import guard | **62, the at-risk set** |
| **at-risk rules exercised by the evidence below** | **62 of 62** |

The 128 taint rules cannot claim a consumer's own type: the receiver must be
tainted from the crate's own constructor, so a locally-declared value never
reaches the sink. Their leak probes measure zero false matches both before and
after, which is what that immunity looks like.

Four of the 62 needed a targeted probe, because they demand structure a generic
probe does not produce — a scoped import (`ark_ec::hashing`, `k256::schnorr`), a
specific argument (`Oaep::new::<Sha256>()`), a value matching a regex
(`Protocol::Tlsv12`), or a FUNCTION-scoped `use`. All four leak before and are
clean after: ark-ec.hash-to-curve 1→0, k256.schnorr-verify 2→0,
native-tls.protocol-version 2→0, rsa.rsa-oaep 2→0.

### False matches removed — leak probes, all 35 exposed crates

One probe per crate: import the crate normally, declare a type the consumer owns
carrying every method name the rules match on a free receiver, and call it in
both spellings the rules use. Every match is a false positive by construction.

| | before | after |
|---|---|---|
| false matches | **174** | **32** |
| crates leaking | **22** of 35 | **3** |

The residual 32 are pkcs8 (18), pkcs1 (6) and sec1 (8), and they are kept
DELIBERATELY: their method names are acute — `to_pkcs8_der`, `from_pkcs1_pem`,
`to_sec1_der` name the crate — so a consumer type carrying one is implementing
that crate's trait. That is the same keep that preserves six true positives in
apple-codesign 0.16.0 and elliptic-curve 0.9.9. The probe is synthetic there; in
real code those matches are correct.

### Nothing lost — 206 published consumers, ~4,500 Rust files

Real reverse-dependencies drawn from the dependency replica, each scanned as one
crate exactly as the miner does, before and after, diffed per `file:line:rule`.

| corpus | consumers | assets before | after | dropped |
|---|---|---|---|---|
| 22 crates with existing corpora | 106 | 789 | 784 | 5 |
| 13 crates pulled for this change | 100 | 877 | 855 | 22 |
| **total** | **206** | **1,666** | **1,639** | **27** |

**All 27 drops were inspected individually and all 27 are correct.** Every one is
a wrong-PURL attribution, not a missed detection:

- `blaze-ssl-async 0.3.7` src/crypto/mod.rs ×6 — declares `mod md5;` and
  `mod sha1;` and implements both by hand. The guard matched the LOCAL MODULE
  NAME, and the md-5 and sha1 crate rules claimed a hand-rolled implementation.
- `wolf-crypto 0.1.0-alpha.15` ×8 — `use wolf_crypto::mac::hmac::{..}`, its own
  wolfSSL-backed `Hmac`, claimed by RustCrypto's hmac rules.
- `mpc-stark 0.2.4` stark_curve.rs:177,178 — `Self::hash_to_field(..)` is a
  private helper in an inherent impl doing a modular reduction, claimed by
  ark-ff's hash-to-field rule.
- `w5500-tls 0.1.3` gcm.rs:23 — `GHash::new(..)` on its own GHASH
  implementation; the guard matched `use super::ghash::GHash`, a local module.
- `pleme-auth-mfa 0.1.2` service.rs:129 — `generator.generate()` on its own
  `BackupCodeGenerator`. Backup codes are not TOTP.
- `apple-codesign 0.16.0` cryptography.rs:332 — `self.try_sign(..)`, the
  wrapper's own dispatch method over an enum that may hold RSA, Ed25519 or
  P-256; attributing that line specifically to p256 is not justified. Line 324,
  which signs on the foreign key-pair type, is retained, so the file keeps its
  ECDSA evidence.

Two earlier versions of this filter DID drop true positives, and the sweep is
what caught both: ten sites lost to trait-implemented methods (fixed by
OwnerTraits and the acute-name keep), then six more lost to an extension trait on
a foreign type (fixed by testing the method rather than its owning type). Neither
was visible to the unit tests.

### Suites

`go test ./...` green, 31 packages. `make lint` 0 issues. `go vet` clean.

One test exists because the first version of this shipped inert: every unit test
passed while production dropped nothing, because the tests used the rule ID as
authored in the YAML and opengrep prefixes it with the rule file's path relative
to the config root — which is how the ruleset is actually loaded.


agustingroh:
Moved to draft — **do not merge yet.**

The published-consumer sweep found a real regression in the filter: it drops 6 true positives.

| crate | assets before | after | lost |
|---|---|---|---|
| pkcs8 | 45 | 40 | 5 |
| pkcs1 | 12 | 11 | 1 |

All six have one cause. The method is declared inside an `impl <CrateTrait> for <LocalType>` block, so the receiver type belongs to the consumer while the method belongs to the crate:

```rust
use pkcs8::{EncodePrivateKey, PrivateKeyDocument};   // pkcs8 trait

impl EncodePrivateKey for InMemoryPrivateKey {        // consumer type
    fn to_pkcs8_der(&self) -> pkcs8::Result<PrivateKeyDocument> { .. }
}
```

Sites: apple-codesign 0.16.0 `src/cryptography.rs:312,313,314,545` (`impl EncodePrivateKey for InMemoryPrivateKey`, trait imported at line 24) and elliptic-curve 0.9.9 `src/secret_key/pkcs8.rs:169` plus one pkcs1 site (`impl FromPrivateKey/ToPrivateKey for SecretKey`).

A method reached through a trait the claimed crate defines IS that crate’s API. The filter needs that fact, and the call graph does not currently carry it: `FunctionDecl.OwnerBases` is empty for Rust even though `processImplBlock` reads the impl header’s trait field.

Fixing that next. The other 11 crates measured in the sweep lose nothing (p256 30→30, k256 40→40, ed25519-dalek 26→26, x25519-dalek 19→19, tokio-rustls 23→23, webpki 1→1, p384 3→3), and the 134→0 false-match result stands.

PR #401, opened 2026-09-02
feat(callgraph): add sequoia-openpgp contracts for the Rust KB

agustingroh:
Adds Rust KB contracts for `sequoia-openpgp` — 24 entries covering certificates
and key material, the streaming `Signer` and `Encryptor`, and the policy-aware
verify and decrypt builders.

Also carries one **separate, pre-existing** lint fix; see the last section.

## Keys read off an exported graph, not from the API

Every key was taken from a `--export-callgraph` run over a probe consumer rather
than written from the crate's documentation. Before this file, all nine keys the
probe emitted appeared as `name(?)` / `name(?, ?)` with empty `parameter_types` —
the shape an absent contract takes. After it, **0 of 9 remain unresolved.**

## `Cert` is contracted under BOTH spellings, and it has to be

The key follows the CONSUMER's import, and this crate re-exports `Cert` at its
root as well as under `cert`:

```
use sequoia_openpgp::cert::Cert;  ->  sequoia_openpgp::cert.Cert.from_bytes
use sequoia_openpgp::Cert;        ->  sequoia_openpgp.Cert.from_bytes
```

Both were measured on probe consumers. A contract carrying one spelling resolves
only half the consumers, and the difference is invisible in the documentation.

## The crate alias is a rule-layer problem, not a contract-layer one

The crate's documentation tells consumers to write `use sequoia_openpgp as
openpgp;`. opengrep does **not** resolve that alias, which is why the rules in
the companion PR carry both spellings. The parser **does**: a probe written
entirely in the aliased form emitted `sequoia_openpgp::…` keys. So one spelling
suffices here, and that asymmetry is recorded in the file rather than left for
the next author to rediscover.

## Version range

`>=0.15.0,<3.0.0`. 0.1.0–0.14.x are excluded deliberately: those releases expose
`tpk` rather than `cert` (renamed at 0.13.0) and have **no `policy` module at
all** — it arrives at 0.15.0 — so the policy-aware entry points contracted here
do not exist there. `Encryptor2` is contracted because it exists inside the
committed range: absent at 1.16.0, present from 1.17.0, gone again at 2.0.0.

## Verification

- Contract test asserting the exact key, arity, role and return for 17 entries,
  plus a **wiring test** in the same commit that asserts the 10 forms an export
  actually emits resolve. `make lint` (pinned golangci-lint v2.10.1): **0 issues**.
  Package tests pass with `-race`.
- Independently confirmed in the family gate: all 7 mined versions serve
  `info_code: READY`, and `/dep-tree` returns 18 crypto entry points for
  `sequoia-openpgp@1.0.0`.

## The pre-existing lint fix, in its own commit

`style(scanner): extract the semgrep package's "unknown" placeholder`

`goconst` fails `make lint` on `internal/scanner/semgrep`: `"unknown"` appears in
`scanner.go`'s version fallback and `transformer.go`'s language sniff. **This
failure predates this branch** — `make lint` on `origin/main` reports the same
single issue, and this branch otherwise touches only the embedded contract KB. It
is a separate commit so it can be reverted independently of the contracts.

---

## Correction (post-merge)

Two counts in this description are wrong. The merged artifact is the authority:

```
$ git show origin/main:internal/callgraph/contracts/rust/sequoia-openpgp.yaml \
    | grep -cE '^\s*-\s*method:'
30
$ ... | grep -oE '^\s*-\s*method:\s*\S+' | sed 's/.*method:\s*//' | sort | uniq -d
sequoia_openpgp::cert::CertBuilder.general_purpose
```

- **30 entries, not 24.** 29 are distinct method keys; the one duplicate is
  `CertBuilder.general_purpose`, present deliberately at both arities because the signature
  changed across the declared range — a contract with only one of them joins for half the
  consumers and silently misses the rest.
- **The contract test asserts 23 entries, not 17** (23 table cases, each checking key, arity, role
  and return).

Neither number changes the contract itself — no commit has touched the YAML since this one
merged — but a description that undercounts what it shipped is worth fixing in place. Found by an
independent review of the downstream bookkeeping change.

### DIFF
--- a/internal/callgraph/contracts/rust/sequoia-openpgp.yaml
+++ b/internal/callgraph/contracts/rust/sequoia-openpgp.yaml
@@ -1,0 +1,223 @@
+schema_version: "2"
+ecosystem: rust
+library:
+  name: sequoia-openpgp
+  coordinates:
+    - sequoia-openpgp
+    - sequoia_openpgp
+  version_range: ">=0.20.0,<3.0.0"
+  description: "Sequoia OpenPGP: certificates, streaming signing/encryption and policy-aware verification"
+
+# API read at sequoia-openpgp 2.3.0, with every pre-2.0 divergence checked against
+# 0.15.0, 0.20.0, 1.0.0 and 1.22.0. Line references name the version they belong to.
+#
+# WHY THE FLOOR IS 0.20.0 AND NOT 0.15.0. The `policy` module arrives at 0.15.0,
+# but the policy-aware ENTRY POINTS contracted here do not: 0.15.0
+# src/parse/stream.rs has no `VerifierBuilder` and no `with_policy` at all — it
+# spells verification `Verifier::from_bytes(policy, bytes, helper, t)`, a
+# four-argument associated function on a different type. The three *Builder types
+# and their `with_policy` first appear at 0.20.0 (src/parse/stream.rs:993, 1402,
+# 1756). 0.15.0 also lacks `Encryptor::for_recipients` (only `for_recipient`,
+# singular, at src/serialize/stream.rs:106) and its `Message::new` returns
+# `writer::Stack`, not `Message`. So 0.15.0 would leave eight of these entries
+# dead, and the range says 0.20.0.
+#
+# WHERE THE KEYS COME FROM. Not from the documentation: each was read off an
+# exported call graph for a probe consumer, and appeared as `name(?)` with empty
+# parameter_types before this file existed.
+#
+# TWO SPELLINGS OF `Cert`, and both are needed. The key follows the CONSUMER's
+# import, and this crate re-exports `Cert` at its root (2.3.0 src/lib.rs:167
+# `pub use cert::Cert;`) as well as under `cert`:
+#
+#   use sequoia_openpgp::cert::Cert;  ->  sequoia_openpgp::cert::Cert.from_bytes
+#   use sequoia_openpgp::Cert;       ->  sequoia_openpgp::Cert.from_bytes
+#
+# Both are authored in the `::` form the rest of this KB uses; the parser emits
+# the receiver separator as a dot and contracts.rustAuthoredKey rewrites it back,
+# so the authored spelling stays conventional.
+#
+# THE CRATE ALIAS IS NOT A PROBLEM AT THIS LAYER. The crate's documentation tells
+# consumers to write `use sequoia_openpgp as openpgp;`. The parser resolves that:
+# a probe written entirely in the aliased form emitted `sequoia_openpgp::…` keys.
+#
+# WHAT THIS FILE CANNOT REACH, measured rather than assumed. The Rust resolver
+# loses a receiver at chain depth 2, so in the fluent form the crate documents —
+# `CertBuilder::new().set_cipher_suite(..).add_signing_subkey().generate()` — only
+# the first call carries a sequoia identity; `add_signing_subkey`,
+# `add_transport_encryption_subkey` and `generate` are attributed to the consumer
+# crate instead. Those three entries are therefore contracted from the source
+# signature, NOT observed in an export, and they will start joining if that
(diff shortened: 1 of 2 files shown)


agustingroh:
## Review response

An independent review agent went over this diff cold, verifying every arity against the
crate's real sources at eight versions. It found real defects; here is what changed and what
did not, with the reason.

### Fixed

| Finding | Fix |
| --- | --- |
| **`general_purpose` had the wrong arity**, so the key joined for no 0.x or 1.x consumer. 0.15.0 `src/cert/builder.rs:151` through 1.22.0 `:377` is `general_purpose<C, U>(ciphersuite, userids)`; only 2.0.0 `:462` drops the first parameter | Both arities are now contracted, with the change point cited. Confirmed independently at 0.15.0, 0.20.0, 1.0.0, 1.22.0, 2.0.0 and 2.3.0 |
| **Six entries were dead at the range floor**, and the stated rationale was wrong: the `policy` module arrives at 0.15.0 but `VerifierBuilder` and `with_policy` do not — 0.15.0 spells verification `Verifier::from_bytes(policy, bytes, helper, t)` | Floor raised to **0.20.0**, where the three builders and their `with_policy` first appear, and the rationale rewritten to say what is actually true. 0.15.0 also lacks `Encryptor::for_recipients` (only the singular) and its `Message::new` returns `writer::Stack` |
| **Return type wrong for 0.20.0–1.22.x**: `Signer.new` and `add_signer` return `Self` there, `Result<Self>` only from 2.0.0 | Documented at the entries with both citations. See below for why it is documented rather than split |
| **The bundled lint commit's premise was false** | Dropped. Verified independently: `make lint` on `origin/main` with a clean cache is `0 issues`, `goconst` needs 3 occurrences and there are 2 non-test ones — and CI reports `Lint: success` on this branch, which is the authority |
| **`Encryptor2` contracted asymmetrically** — factory only, so a consumer pinned to 1.17–1.22 lost both config calls | `add_recipients` and `symmetric_algo` added. Also added `with_passwords`, `with_session_key` (`stream.rs:2761`) and `aead_algo` (`:3004`), all verified present at the 0.20.0 floor |
| **The `Cert` keys were the only 3-segment dot-form keys in the whole Rust KB** | Both authored in the conventional `::` form; `rustAuthoredKey` rewrites the emitted dot form onto them. Comment corrected so it no longer presents the dot form as required |
| **`with_policy` role inconsistent** with sibling factories | Changed to `factory`: it returns the Verifier/Decryptor and does no cryptography at the call site |
| **The "0 of 9" headline was achieved by choosing the probe** | Restated. On a probe covering the whole contracted surface: 14 keys emitted, **2 unresolved** — and those two are attributed to the *consumer* crate, not to this one, by the resolver's depth-2 receiver loss |

30 entries now. Contract and wiring tests pass, `-race` clean, `make lint` `0 issues`.

### Not fixed, and why

- **The pre-2.0 return type is documented rather than contracted.** The schema allows one
  return per key + arity, and both `Self` (0.20.0–1.22.x) and `Result<Self>` (2.x) occur inside
  the range. The 2.x form is declared because it is the current stable line and what the mined
  sample uses; the divergence is recorded at the entry with both citations so a reader is not
  misled. Splitting it is not expressible without changing the schema.
- **`CertBuilder.generate`, `add_signing_subkey` and `add_transport_encryption_subkey` never
  join.** Correct, and now said plainly in the file: the Rust resolver loses the receiver at
  chain depth 2, so in the fluent form the crate documents only the first call carries a
  sequoia identity and the rest are attributed to the consumer crate. Those three are
  contracted from the source signature, **not** observed in an export, and the file says so —
  they will begin joining if that resolver limit is lifted. Anchoring elsewhere would not help;
  this is an engine limitation, not a key typo.
- **`version_range` is not consumed for filtering.** Noted and left as is — it is metadata for
  downstream range consumers, which is exactly why getting the floor right mattered.

PR #407, opened 2026-09-02
feat(callgraph): add schannel contract knowledge base

agustingroh:
Adds a callgraph contract KB for the `schannel` crate — the Rust binding to Windows SChannel and
CryptoAPI/CNG — so a consumer's TLS handshake, credential acquisition, certificate parsing,
fingerprinting and private-key import call sites resolve to a declared signature instead of an
empty one. 56 entries.

## Why the config entries exist, which is the point of the file

schannel's API is fluent builders end to end, so almost every interesting call sits at hop 2 or
later of a chain. Without a declared return type for each link, those hops resolve to the
**scanned crate's own name** — a *wrong* identity rather than a missing one, because it looks
resolved and matches no contract. Measured on a realistic consumer at three points:

| State | Call sites resolved |
| --- | --- |
| before this contract existed | **5 of 12** |
| with the `role: config` links declared | **9 of 12** |
| with `canonical_return_type` added | **12 of 12** |

`domain`, `container`, `new_keyset` and the `ProviderType`/`HashAlgorithm` constructors are not
crypto operations and are not reported as any. They are the links a chain travels to reach one,
the same way `boring.yaml` catalogues `MessageDigest.sha1` and `Cipher.aes_128_gcm`.

**The concrete type goes in `return`, the wrapper in `canonical_return_type`**, following
`security-framework.yaml`. That is load-bearing, not cosmetic: with `return: core::result::Result`
the chain `prov.import().import(der)` resolved to `core::result.Result.import`, because
`.unwrap()` yielded the wrapper and the next link was looked up on it.

## Four entries are declared on a type that does not have those methods

`SchannelCred.enabled_protocols`, `.supported_algorithms`, `.cert` and `.acquire` are declared
because that is the identity the resolver **emits**, and a contract keyed on the truthful type
joins nothing for the crate's own idiomatic entry point. An associated function's return type is
taken from the path whose function was called — right for `Type::new()`, wrong for a builder
accessor — so `SchannelCred::builder()` yields `SchannelCred` even though this file declares its
return as `schannel_cred::Builder`.

The truthful `Builder.*` entries are kept alongside, and the note in the file says to delete the
four aliases once the resolver takes an associated function's return from its contract. Precedent:
`sequoia-openpgp` contracts `Cert` under both spellings.

## The version floor is 0.1.8

An earlier revision of this PR said **0.1.10** and justified it with `crypt_prov`, `crypt_key`,
`key_handle` and `ncrypt_key` together. Three of those four were wrong to cite:

```
$ for v in 0.1.1 0.1.7 0.1.8; do echo -n "$v: "; ls $v/src/ | tr '\n' ' '; echo; done
0.1.1: cert_context.rs cert_store.rs context_buffer.rs ctl_context.rs key_handle.rs lib.rs …
0.1.7: cert_context.rs cert_store.rs context_buffer.rs ctl_context.rs key_handle.rs lib.rs …
0.1.8: cert_chain.rs cert_context.rs cert_store.rs … crypt_key.rs crypt_prov.rs key_handle.rs
       … ncrypt_key.rs schannel_cred.rs …
```

`crypt_key` and `ncrypt_key` arrive at **0.1.8** alongside `crypt_prov`; `key_handle` exists as
early as **0.1.1**; and this contract declares **no method** on any of the three — `crypt_key`
appears only as a return type. So the floor was two releases too high and three quarters of its
stated reason was vacuous. It is set by `crypt_prov` and nothing else. Also corrected: `0.0.2` is
a second release of the pre-API crate, not only `0.0.1`.

**Four entries post-date the floor** and are now annotated at their declarations rather than left
to look as though a single check covered them: `use_sni` and `accept_invalid_hostnames` arrive at
0.1.12, `TlsStream.peer_certificate` at 0.1.13, `negotiated_application_protocol` at 0.1.18.

**`version_range` has no reader in this repository** — `Load()` parses it into
`Library.VersionRange` and nothing consults the field — so the floor is provenance for a human
rather than a match-time filter. Worth getting right for that reason alone. Splitting into two
libraries with different ranges is not expressible: `yamlContract` has no per-entry version field,
and `validateLibraryUniqueness` rejects two files declaring the same library name.

## The two crypto-determining calls are no longer plumbing

`supported_algorithms` and `enabled_protocols` select the cipher suite and the TLS version. They
now carry `parameters:` with `contributes: { property: algorithm | protocolVersion, derivation:
argument_type }`, so the selection is derivable instead of sitting under the same bare `config`
label as `silent` and `machine_keyset`. Both take a slice, so `argument_type` names what the
argument's *type* says — all this layer can prove without evaluating the literal.

## Two silently partial sets, completed

- All **ten** `ProviderType` constructors. Three were omitted with no note, which reads as a
  judgement rather than the oversight it was.
- The private-key chain's **terminal operation**: `AcquirePrivateKeyOptions.acquire` plus its two
  links. Declaring `CertContext.private_key` and stopping left the one chain this file's own
  rationale describes with a root, no links and no operation.

## The tests run through the full builder

`rust_fluent_chain_qualifier_test.go` records why: *"the parser resolved the case above correctly
and a post-build pass undid it, so a parser-only helper asserts nothing about this defect."* Two
passes run after parsing, and the first version of these tests saw neither.

Three tests, each checked against a mutation rather than assumed to work:

1. **Every identity the builder emits joins exactly one contract** with the expected role and
   library — and the set of schannel keys in the KB must be either exercised or explicitly listed
   as reachable-but-unexercised, so an added contract cannot pass unnoticed.
2. **No call site in a realistic consumer may fall back to the scanned crate.** This is the
   assertion that would have caught the state this KB started in, where 7 of 12 sites came out
   owned by the consumer's own package.
3. **Every declared arity is asserted against the crate's real signatures.** This one exists
   because arity is *documentation* for Rust: callees carry no encoded arity, every lookup arrives
   with `-1`, and `ContractsFor` resolves by name — so mutating `Builder.connect` from 2 to 7
   changes no match and is invisible to tests 1 and 2.

All four mutations a review used to show the first version of these tests was inert — delete an
entry, flip a role, change an arity, append a contract for a module that does not exist — now
fail.

## Verification

```
$ GOPRIVATE='github.com/scanoss/*' go test -count=1 ./...   -> 31 ok, 0 FAIL
$ ./bin/golangci-lint run   (clean GOLANGCI_LINT_CACHE + GOCACHE)   -> 0 issues.
```

The clean cache matters: a warm cache reported one `goconst` hit in a file this branch does not
touch, and `origin/main` with the same warm cache reported none — so the hit was cache state, not
this change. Verified rather than assumed to be pre-existing.

## Related

The `ChainID` gap reported in #408 is real and unchanged — the Rust parser sets no `ChainID`, so
`resolveFluentChainCalleesByContract` cannot run for Rust. It is **not** the operative cause of
what this KB works around, and #408 has been corrected accordingly.


agustingroh:
## Review response

An independent agent reviewed this branch cold. Nine of its twelve findings were real; two were
correct at the commit reviewed and have since been superseded by a change on `main`; one I
disagree with and say why. Everything is in `7513a00`.

### Fixed

**1, 3 — the version floor was wrong by two releases, and three quarters of its justification was
vacuous.** Confirmed by reading `ls src/` at each version rather than re-reading my own table:
`crypt_prov`, `crypt_key` and `ncrypt_key` all arrive at **0.1.8**, and `key_handle` exists at
0.1.1. And your sharper point stands — this contract declares no `method:` on `crypt_key`,
`key_handle` or `ncrypt_key` at all, so citing them was meaningless. The floor is 0.1.8, set by
`crypt_prov` and nothing else. Also corrected: `0.0.2` is a second release of the pre-API crate,
which the text named only as `0.0.1`.

**2 — four entries post-date the floor and the header implied a check that had not happened.**
Confirmed: `use_sni` and `accept_invalid_hostnames` at 0.1.12, `TlsStream.peer_certificate` at
0.1.13, `negotiated_application_protocol` at 0.1.18. Each is now annotated at its declaration.

**4 — 34 of 46 entries unguarded; four hostile mutations passed.** The most valuable finding here.
Fixed in two parts, and I reproduced your mutations to check:

- an exhaustiveness pass over `kb.Contracts` requires every schannel key to be either exercised
  by the test source or explicitly listed as reachable-but-unexercised, so your fabricated
  `schannel::totally_fake::NotAThing.definitely_not_real` now fails, as do the deletion and the
  role flip;
- your arity mutation (2 → 7) still passed all of that, for exactly the reason you identified —
  Rust lookups arrive with `-1` and resolve by name, so arity is documentation. It now has its own
  test asserting every declared arity against the crate's real signatures, and the mutation fails
  there.

**6 — the tests were parser-only, and the repo's own helper warns against it.** Confirmed;
`rust_fluent_chain_qualifier_test.go`'s comment is unambiguous. Both tests now go through
`BuildFromDirectories`.

**9 — the crypto-determining calls were buried as untyped `config`.** Agreed and fixed.
`supported_algorithms` and `enabled_protocols` carry `parameters:` with
`contributes: { property: algorithm | protocolVersion, derivation: argument_type }`, following
`openssl.yaml`'s shape. Your framing — the crypto-relevant entries under-modelled while the
trivial ones were over-modelled — was the right diagnosis.

**10 — incomplete sets.** All ten `ProviderType` constructors now (the three omitted had no note,
which read as a judgement rather than the oversight it was), and the private-key chain's terminal
operation `AcquirePrivateKeyOptions.acquire` plus its two links. You put it best: the file's own
rationale is that config links exist so a walk can reach an operation, and that chain had a root,
no links and no operation.

**11, 12 — the mis-keyed escape valve and "the matrix".** The test was rewritten so the guard is
moot; "the matrix" is gone from the public text.

**8 — `version_range` has no reader in this repository.** Confirmed, and now stated in the header,
so the floor is presented as provenance rather than as a filter. Your sub-item is right too and I
kept the claim: splitting is genuinely not expressible — no per-entry version field, and
`validateLibraryUniqueness` rejects two files with the same library name.

### Superseded rather than wrong-at-the-time

**7 — the central diagnosis.** Your verification of it is correct: the Rust parser sets no
`ChainID`, so `resolveFluentChainCalleesByContract` is inert for Rust. But the *consequence* I drew
from it — that the `role: config` entries are correct-but-unused — is no longer true, because
`contractReturnType` landed in `rust_type_semantics.go` with #403 while this branch was open. It
resolves a method call on a third-party receiver through the KB, which is exactly what makes those
entries work. Measured on the same source at three points: **5 of 12 call sites resolved** before
the contract, **9 of 12** with the config links, **12 of 12** once `canonical_return_type` carried
the concrete type behind each `io::Result<T>`.

That last step is your finding 5 in another guise: with a bare `core::result::Result` return, the
chain `prov.import().import(der)` resolved to `core::result.Result.import`, because `.unwrap()`
yielded the wrapper. #408 is corrected accordingly, and the framing you objected to — "two
independent causes", "neither fixable from the YAML" — is withdrawn there as well. You were right
that cause 1 is subsumed by cause 2, and right that alias entries were available.

**5 — `SchannelCred.enabled_protocols` joins zero contracts.** Confirmed at the commit you
reviewed, and now fixed the way you suggested: four alias entries declared under the identity the
resolver actually emits, with the `sequoia-openpgp` precedent cited and a note saying to delete
them once an associated function's return type comes from its contract. Every call site in a
realistic consumer now resolves to schannel — a test asserts that none falls back to the scanned
crate, which is the assertion that would have caught the original state.

### On the build-health note

Your `make lint` → `0 issues` matches mine **with a clean cache**. A warm shared
`GOLANGCI_LINT_CACHE` reported one `goconst` hit in a file this branch does not touch, and
`origin/main` with the same warm cache reported none — so the hit was cache state. I verified that
rather than reporting it as pre-existing, which is a mistake I have made before on this campaign.

`go test ./...` 31 ok, 0 FAIL. Contract entries: 56.

PR #409, opened 2026-09-02
feat(callgraph): add aes-kw contract knowledge base

agustingroh:
Types the `aes-kw` (AES Key Wrap, RFC 3394 / RFC 5649) call sites so a consumer's wrap and unwrap operations resolve instead of coming back with an empty parameter list.

Keys were read off an **exported call graph**, not assumed.

## Both API lines are recorded, because both are published and mineable

| versions | API |
|---|---|
| 0.1.0 – 0.2.1 | `Kek<Aes>` + `KekAes128/192/256`; `wrap`/`unwrap` (arity 2), `wrap_vec`/`unwrap_vec` (arity 1), and from 0.2.0 the padded forms. |
| 0.3.0 – 0.3.1 | `Kek` is **gone**. `AesKw<C>` / `AesKwp<C>` + `KwAes*` / `KwpAes*`; `wrap_key`/`unwrap_key` (arity 2), `wrap_fixed_key`/`unwrap_fixed_key` (arity 1). |

`0.0.0` is outside the declared range on purpose: it is a yanked, empty placeholder whose `src/lib.rs` is 0 bytes, so it declares nothing to type.

## Why this is 76 entries and not 19

**The parser keys on the alias the consumer wrote.** A consumer calling `KwAes128::new(..)` then `kw.wrap_key(..)` emits

```
aes_kw.KwAes128.new        aes_kw.KwAes128.wrap_key
```

and never `aes_kw.AesKw.*`. The aliases are declared *inside* the crate (`pub type KwAes128 = AesKw<aes::Aes128>;`), so nothing in the consumer's source says what they expand to, and an entry on the generic type alone would load cleanly and resolve nothing. `rust/aes-gcm.yaml` enumerates `Aes128Gcm` / `Aes256Gcm` for the same reason. The generic names are recorded too, for `AesKw::<Aes128>::new(..)` and the older bare `Kek`.

`TestAesKwEveryAliasIsTyped` asserts every alias of both lines carries its own entries, with the arity, role, parameter-type count and return type of each.

## Arity keying is what keeps `Result::unwrap` out

Measured: `kek.wrap(data, &mut out).unwrap()` can emit `aes_kw.KekAes128.unwrap` with **arity 0** — the receiver routing follows the chain back to `kek`, so the trailing `Result::unwrap` is attributed to the crate's type. The crate's own `Kek::unwrap(&self, data, out)` is arity 2.

`TestAesKwHasNoZeroArityUnwrap` asserts no alias carries a zero-arity `unwrap`, and that the genuine two-argument form still resolves. If contracts ever stop keying on arity, this becomes a wrong typing on the single most common method call in Rust.

## Verification

The wiring tests were **mutation-checked**: deleting all nine `KekAes192` entries, deleting the whole `AesKw` line, flipping a role, and adding a zero-arity `unwrap` each make them fail. (The first two initially survived — which turned out to be because an earlier edit of mine had silently truncated two of the tests. The mutation pass is what surfaced it.)

`make lint` with the repo's pinned golangci-lint: 0 issues. Full suite unchanged from `origin/main`: 31 packages ok, 0 failures.

### DIFF
--- a/internal/callgraph/contracts/rust/aes-kw.yaml
+++ b/internal/callgraph/contracts/rust/aes-kw.yaml
@@ -1,0 +1,593 @@
+schema_version: "2"
+ecosystem: rust
+library:
+  name: aes-kw
+  coordinates:
+    - aes-kw
+    - aes_kw
+  version_range: ">=0.1.0,<0.4.0"
+  description: "AES Key Wrap (RFC 3394) and Key Wrap with Padding (RFC 5649)"
+
+# API read at 0.3.1 (src/kw.rs, src/kwp.rs, src/lib.rs) and at 0.2.1 and 0.1.0
+# (src/lib.rs; 0.1.0 keeps it in src/kek.rs). 0.2.0 and 0.3.0 were diffed against 0.2.1 and 0.3.1:
+# every `pub fn` and `pub type` signature is identical, so no arity drifts
+# inside either line.
+#
+# TWO INCOMPATIBLE APIS INSIDE THE DECLARED RANGE, both recorded because both
+# are published and mineable:
+#
+#   0.1.0-0.2.1  `Kek<Aes>` with the `KekAes128/192/256` aliases. Constructed by
+#                `new`, `From` and `TryFrom`. Methods `wrap`/`unwrap` (arity 2)
+#                and `wrap_vec`/`unwrap_vec` (arity 1).
+#   0.3.0-0.3.1  `Kek` is GONE. `AesKw<C>` / `AesKwp<C>` with the
+#                `KwAes128/192/256` and `KwpAes128/192/256` aliases, and methods
+#                `wrap_key`/`unwrap_key` (arity 2) and
+#                `wrap_fixed_key`/`unwrap_fixed_key` (arity 1).
+#
+# THE PADDED FORMS ARRIVE AT 0.2.0. `wrap_with_padding`, its `_vec` form and
+# both unwrap counterparts do not exist in 0.1.0; the entries cover the rest of
+# the range.
+#
+# `new` IS A TRAIT METHOD IN 0.3.x. The 0.3.0 changelog removed the inherent
+# `Kek::new` in favour of implementing `InnerInit`, so `new` now arrives through
+# the blanket `KeyInit` impl that lib.rs re-exports. Arity 1 either way, and the
+# key is what the consumer writes, so the entries are unaffected.
+#
+# `from` DECLARES NO parameter_types, AND THAT IS DELIBERATE. `From` is an
+# OVERLOAD SET here: 0.2.1 lib.rs:95-129 declares six impls, two per alias at
+# arity 1 -- `From<GenericArray<u8, U16|U24|U32>>` and `From<[u8; 16|24|32]>` --
+# and none of them is generic over `Aes`. AGENTS.md says to omit the canonical
+# fields when several same-arity overloads remain possible, so call-site
+# provenance selects the signature instead of this file guessing one. An earlier
+# revision recorded `GenericArray<u8, Aes::KeySize>`, which is the `new`
+# signature's spelling, appears nowhere among the `From` impls, and named the
+# overload the crate's own 12 call sites never use (they all pass `hex!(..)`,
+# i.e. `[u8; N]`).
+#
+# `From` AND `TryFrom` ARE THE IDIOMATIC CONSTRUCTORS ON THE 0.1-0.2 LINE, and
+# leaving them out left that whole factory path untyped. The crate's README and
+# every one of its own tests construct with `Kek::<Aes128>::from(..)` or
+# `try_from(..)` and never with `new`. Measured before they were added: those
+# call sites emitted `aes_kw.Kek.from` / `aes_kw.Kek.try_from` and resolved to
+# nothing. 0.3.x implements neither, so the entries stop at the Kek line.
+#
+# EVERY ALIAS IS ENUMERATED, and that is not padding. The keys were read off an
+# exported call graph, and the parser keys on the alias the CONSUMER wrote: a
+# consumer calling `KwAes128::new(..)` emits `aes_kw.KwAes128.new`, never
+# `aes_kw.AesKw.new`. The aliases are declared inside the crate
+# (`pub type KwAes128 = AesKw<aes::Aes128>;`), so nothing in the consumer's
+# source says what they expand to. rust/aes-gcm.yaml enumerates
+# `Aes128Gcm` / `Aes256Gcm` for the same reason.
+#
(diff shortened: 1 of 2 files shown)


agustingroh:
## Review findings addressed — 76 entries → 84, and the test was hollow where it mattered

An independent review found five real defects. All are fixed, and two of them changed what this PR claims.

**1. `From` and `TryFrom` were missing — the whole 0.1–0.2 factory path was untyped.** The contract had `Kek.new` only, but the crate's README and **every one of its own tests** construct with `Kek::<Aes128>::from(..)` or `try_from(..)` and never with `new` — verified: 12 `from` and 2 `try_from` call sites across the 0.2.1 tests, zero `new`. Those emitted `aes_kw.Kek.from` / `aes_kw.Kek.try_from` and resolved to nothing. **Fixed**: 8 entries added across the four Kek-line types. 0.3.x implements neither, so they stop at that line. My own wiring test missed this because its synthetic consumer used the one form the crate itself does not.

**2. `KwWrappedKey` / `KwpWrappedKey` are not nameable by a consumer.** `mod kw;` and `mod kwp;` are private (lib.rs:16-17) and lib.rs re-exports only `AesKw`, `AesKwp` and the error types — `grep WrappedKey src/lib.rs` returns nothing. Eight entries named `aes_kw::KwWrappedKey`, a path that does not exist. **Fixed**: the real paths `aes_kw::kw::KwWrappedKey` / `aes_kw::kwp::KwpWrappedKey`, with the situation stated in the header.

**3. `unwrap_fixed_key`'s parameter type was wrong.** It takes `&KwWrappedKey<N>` = `&Array<u8, Sum<N, IvLen>>` (kw.rs:213-216), a different width from `wrap_fixed_key`'s `&Array<u8, N>`. **Fixed** on all 8 entries.

**4. The arity claim was measured wrong, and I have corrected it rather than restated it.** The comment said `kek.wrap(data, &mut out).unwrap()` emits `aes_kw.KekAes128.unwrap` at arity 0. It does not — that yields `core::result.Result.unwrap`. The trigger is `kek.unwrap(data, &mut out).unwrap()`, where the **inner** method shares the name. Re-measured and rewritten.

   The review also found a **latent second hole** I had not seen: Rust callee names carry no `#N` suffix, so `splitMethodArity` returns -1 for every aes-kw call and `rustContractsFor` falls through to `lowestArityByName`. Confirmed: `ContractsFor("aes_kw.KekAes128.unwrap", -1)` returns the crate's `unwrap/2`. Every path I traced passes a real `len(call.Arguments)`, so it is not a live wrong typing today. It is now documented and asserted as current behaviour rather than left implicit.

**5. `keySize` metadata was missing**, against a very consistent convention — `aes.yaml`, `aes-gcm.yaml`, `blowfish`, `des`, `cbc`, `ctr`, `hmac` and others all attach `contributes: {property: keySize, derivation: argument_bit_length}` to their key-taking factory. For a library whose twelve aliases are *defined* by key size, dropping it was the wrong call. **Fixed** on every factory.

## And the test was hollow exactly where these got through

The review's mutation sweep killed all 76 single-entry deletions, role flips and arity changes — but **five mutations survived**: a bogus return type, a factory returning another type, garbage parameter-type strings, a KW/KWP return swap, and a confidence downgrade. Cause: the test asserted only that the return type was *non-empty*, and that the parameter-type *count* equalled the arity — which the loader already validates, so that line could never fail on its own.

Return type and parameter types are the two fields that encode the API, and findings 2 and 3 are precisely what that gap let through.

**Fixed**: the test now asserts the exact return type, the exact parameter-type strings, the role, and the `keySize` contribution for every one of the 84 entries. I re-ran all five surviving mutations: **all five now fail.**

`make lint` 0 issues; 31 packages ok, 0 failures.

agustingroh:
## Second review round — three more fixed, and a claim of mine corrected again

**`.from` was guessing one overload out of six.** `From` is an overload set here: 0.2.1 lib.rs:95-129 declares six impls, two per alias at arity 1 — `From<GenericArray<u8, U16|U24|U32>>` and `From<[u8; N]>` — and **none is generic over `Aes`**. AGENTS.md says to omit the canonical fields when several same-arity overloads remain possible. Worse, the spelling I recorded (`GenericArray<u8, Aes::KeySize>`) is the `new` signature's, appears nowhere among the `From` impls, and named the overload the crate's own twelve call sites never use — they all pass `hex!(..)`, i.e. `[u8; N]`. **Fixed**: `parameter_types` dropped from all four `.from` entries, with the reason in the header.

**`confidence: high → low` still survived on all 84 entries.** My previous comment said "all five now fail" — that was wrong; four died, this one did not. It flows into `InferredReturn.Confidence` and is asserted in 36 other test files, so an unasserted downgrade is a silent behaviour change. **Fixed**: the confidence is now asserted per entry.

**The keySize assertion was below the bar of the guard it cites.** `aes.yaml`'s guard asserts Role + Property + Derivation, and `rust_libraries_test.go` asserts the index; mine checked only the property, so index, name, parameter role and derivation all mutated freely. **Fixed**: all five are asserted, extracted into a helper so the nesting stays within lint limits. Re-ran the sweep — `confidence→low`, `derivation→argument_value`, `index 0→1`, `name key→nonce` and `role→operation-determining` all now fail.

**And my arity explanation was wrong for the second time.** I had narrowed it to "only when the inner method shares the name". Also false. Measured:

```
kek.wrap(d, &mut o).unwrap()          -> core::result.Result.unwrap   n=0
KekAes128::try_from(k).unwrap()       -> aes_kw.KekAes128.unwrap      n=0
KekAes128::try_from(k).expect("bad")  -> aes_kw.KekAes128.expect      n=1
```

The trigger is the Result-extraction routing on a **path-qualified associated-function receiver** — the crate's type carries onto whatever extracts the Result, `unwrap` or `expect` alike. That shape is the crate's own idiom, and adding the `try_from` entries is what made it reachable, so the guard matters more after that change rather than less.

I had also written that "every path passes a real `len(call.Arguments)`". Not accurate: `inference.go`'s `candidateFromCallResult` passes `splitMethodArity`'s raw -1, and `contracts.go`'s own doc names it as that caller. What actually prevents a wrong typing is `shouldInfer` suppressing a function with a concrete declared return — verified on `fn make(k: &[u8]) -> KekAes128 { KekAes128::try_from(k).unwrap() }`: inferred nil, suppressed. The comment now says that.

Also corrected the header nit: 0.2.1 has no `src/kek.rs` — that is 0.1.0's layout.

**Not changed, and why**: the wiring test's synthetic consumer still constructs with `::new`. The review is right that this is the same hollowness that hid the missing `from`, and I verified end-to-end that `from` and `try_from` do resolve through the parser — but the *regression guard* is still the KB row, not the parse. I have left it rather than widen the diff further; it is worth a follow-up.

`make lint` 0 issues; 31 packages ok, 0 failures.

PR #413, opened 2026-09-02
feat(callgraph): add bcrypt contract knowledge base

agustingroh:
Adds a callgraph contract knowledge base for the `bcrypt` crate, so a consumer's password hashing and verification call sites resolve to a declared signature instead of an empty one.

### Keys

The crate's callable surface is free functions in a single-segment module (the `HashParts` accessors and the `Version` enum are public too, and are listed as skipped), so every key is the `bcrypt.<fn>` dot form with no module segment. The keys were read off an exported callgraph rather than written from the documentation: a probe consumer scanned with `--export-callgraph` emitted `bcrypt.hash`, `bcrypt.verify`, `bcrypt.hash_with_salt`, `bcrypt.non_truncating_hash`, `bcrypt.non_truncating_verify` and `bcrypt.bcrypt` at the arities declared here.

An external call renders as `name(?, ?)` with empty `parameter_types` in that export whether or not a contract exists — a resolved `scrypt.scrypt` call renders identically — so `rust_bcrypt_contract_wiring_test.go`, not the rendering, is what pins these keys. It parses the two spellings a consumer writes (a named import and the crate path) and asserts `ContractsFor(key, arity)` returns exactly one `bcrypt` contract for each.

### Two decisions worth reviewing

**Cost is declared as an argument, never encoded.** Every hashing function takes the cost at the call site, so the contract names the parameter and the finder derives the value; nothing here claims a constant. The one source-proven value, `DEFAULT_COST = 12` (unchanged from 0.1.0 through 0.19.2), belongs to the detection rules.

**The raw Eksblowfish primitive is declared at both of its arities.** `bcrypt::bcrypt` wrote its digest into a caller-supplied buffer through 0.12.1 (`cost, salt, password, output`) and returns `[u8; 24]` from 0.13.0 (`cost, salt, password`). Recording only the modern arity would leave every pre-0.13 call site unresolved while looking complete, so both are contracted and `TestBcryptRawPrimitiveIsContractedAtBothArities` pins that.

`version_range` spans `>=0.1.0,<0.20.0`, but only `hash` and `verify` exist across all of it — the other twelve entries begin at the releases below, and a contract for a spelling that does not yet exist matches nothing. Signatures read at 0.19.2 (`src/lib.rs`, `src/bcrypt.rs`); earlier spellings read at the release that made each **reachable by a consumer**, which is not always the release its code appeared in — `hash`/`verify` 0.1.0, the raw primitive **0.5.0** (`src/bcrypt.rs` exists from 0.2.0 but `lib.rs` keeps the module private until then), `hash_with_result` 0.6.0, `hash_with_salt` **0.6.3**, the `non_truncating_*` family 0.16.0, the `*_bytes` variants 0.19.0.

Two further changes are type-only with unchanged arity — `hash_with_salt`'s salt `&[u8]` → `[u8; 16]` at 0.13.0, and `hash`/`verify`'s password `&str` → `P: AsRef<[u8]>` at 0.2.2 — so they need no second entry.

### Declared signatures

**Eleven of the fourteen entries declare `parameter_types` and `canonical_return_type`.** That is what fills the exported canonical signature: without them the export carries `bcrypt.hash(?, ?)` and empty parameter types, which is the very thing this KB exists to replace. An earlier revision omitted them all, on the mistaken reasoning that an external call renders as `(?, ?)` regardless — a comparison drawn against `scrypt`, whose contract happens not to declare them either.

The three that still omit them are the three where method plus arity does not identify one signature across the range: `hash` and `verify` (password type changed at 0.2.2) and `hash_with_salt` (salt type changed at 0.13.0). Declaring either era's types there would be right for half the range and wrong for the other half.

`DEFAULT_COST`, `BASE_64` and `Version` are not call sites and are deliberately absent; `TestBcryptConstantsAreNotContracted` pins that a callable identity for them never enters the KB. `encode`/`decode` (0.1.0–0.7.0) and the `HashParts` accessors are base64 and formatting helpers over an already-computed hash.

### Validation

- `go test -race ./...` — 31 packages ok, 0 failures, no data race; same as `origin/main`.
- `make lint` at the repository's pinned golangci-lint — 0 issues.
- The wiring tests cover **all 14** entries, each at its declared arity and at every neighboring arity from 0 to 5, plus each entry's return type and the property and derivation of every parameter, and they assert the KB holds exactly 14 bcrypt entries. Verified by mutation, not by inspection: swapping `password` and `salt`, renaming a key to a spelling no parser can emit, deleting the arity-4 entry, reverting a return type, and adding a spurious fifteenth entry each fail the suite.


agustingroh:
Independent review of this PR found six real defects. All six are fixed in the amended commit; here is what each was and how it was verified, since several are not visible from the diff alone.

**1. The `output` parameter claimed a length that can never resolve.** The pre-0.13 raw primitive writes into `output: &mut [u8]`, and a slice type carries no length, so `{property: dataLength, derivation: argument_type}` could never yield a value — and it broke convention twice over: `dataLength` is paired with `argument_value` in all 25 other uses in the KB, and `argument_type` is used elsewhere only to derive a semantic label from a type name. Now `{property: output, derivation: argument_value}`, matching `argon2`'s caller-supplied buffer, which is the exact analogue already in this KB.

**2. A stored bcrypt digest was declared as `ciphertext`.** `ciphertext` has 84 uses in the KB, all for genuinely encrypted data. The house spelling for a stored password hash is `passwordHash` — `go/golang-x-crypto.yaml`'s `bcrypt.CompareHashAndPassword` index 0, the same argument of the same algorithm. Both `verify` and `non_truncating_verify` now use it. Left as it was, every password-verification site would have been conflated with encryption findings downstream.

**3. `return: {type: none}` was an invented token.** `none` appeared exactly once in the entire contracts tree: this file. Rust's unit type here is `()` (170 uses). Fixed.

**4. A false provenance claim in the file header.** It said the raw primitive arrived at 0.2.0. `src/bcrypt.rs` does exist from 0.2.0, but `lib.rs` declares `mod bcrypt;` **private** through 0.4.0; the first `pub use crate::bcrypt::bcrypt;` is in **0.5.0**. So no consumer could call it before 0.5.0, and the header was documenting a private internal function as public API. Verified by reading `lib.rs` at 0.2.0, 0.3.0, 0.4.0, 0.5.0 and 0.6.0. The same error appeared in the skipped-API list: `encode`/`decode` live in `mod b64`, which is private at **every** published version, so they were never consumer-reachable at all. Both corrected, and the header now states the rule it got wrong — a spelling's provenance is the release that made it reachable, not the release its code appeared in.

**5 and 6. The tests did not protect what was most likely to be wrong.** The review demonstrated this by mutation rather than by argument: swapping the `password` and `salt` properties on `hash_with_salt` left the whole suite green, and so did renaming two keys the wiring test did not exercise — 8 of the 14 entries rested on nothing.

Both are closed, and I verified the closure the same way, by mutation:

| mutation | before | now |
| --- | --- | --- |
| swap `password` / `salt` on `hash_with_salt` | green | **fails** |
| rename `bcrypt.hash_bytes` → `bcrypt.hashBytes` | green | **fails** |
| delete the arity-4 raw entry | fails | **fails** |

`TestBcryptEveryContractedIdentityResolves` now looks up all 14 entries at their declared arity **and** at every neighboring arity from 0 to 5, so an arity that is ignored rather than matched is caught too. `TestBcryptParameterSemantics` asserts the property and derivation of every parameter of every entry, in the house style of `rust_password_hashes_test.go` — including the raw primitive's argument order, which is `cost, salt, password` and not the `password, cost` order every other function uses.

**Also fixed:** `TestBcryptConstantsAreNotContracted` would have passed with `bcrypt.yaml` deleted entirely. It now asserts the KB is loaded before asserting what is absent, and probes arities 0–4 rather than 0–1.

**One thing I removed rather than defended:** the PR body claimed end-to-end mining validation that nothing in this PR lets a reader check. It is true, but it is evidence from a private tracker and it does not belong in a public PR body as an unverifiable assertion. The verifiable claims — the mutation results above, `go test -race ./...` (31 packages, no race), `make lint` at the pinned version (0 issues) — stand on their own.

agustingroh:
Second review round found one substantive test gap and four provenance errors. All fixed.

### The gap was exactly where the previous round's fix lived

**No test read `Contract.Return`.** So reverting the unit type from `()` back to `none` — the precise defect round one found and fixed — left the suite green, and so did changing `hash`'s return to `bool`. A fix with no assertion behind it is a fix waiting to be undone.

`TestBcryptEveryContractedIdentityResolves` now carries a return type per entry and checks it. Verified by mutation:

| mutation | before | now |
| --- | --- | --- |
| `return: {type: ()}` → `{type: none}` on the arity-4 raw entry | green | **fails** |
| `hash` returns `bool` instead of `Result` | green | **fails** |
| a spurious 15th entry (`bcrypt.encode`) added | green | **fails** |

The third came from the review's observation that the test counted *expected* entries found, so it could not notice an entry that should not exist. It now also counts what the KB actually holds for this library and requires exactly 14.

### Provenance, corrected

- **`hash_with_salt` is 0.6.3, not 0.7.0.** Absent from 0.6.2, present from 0.6.3 — published nine days before 0.7.0. The independent review of the rules PR found the same error independently, which is why I have also added 0.6.3 to the mined sample rather than only fixing the comment.
- **The arity-4 raw entry's reachable range is 0.5.0–0.12.1, not 0.2.0–0.12.1.** Round one corrected the header to say provenance is the release that makes a spelling reachable, and then left this comment contradicting it two hundred lines lower.
- **`mod b64` is not "private at every published version"** — it exists at 0.1.0–0.7.0 and from 0.8.0 the module is gone entirely. The conclusion is right, the evidence for it was wrong for 32 of 39 versions.
- **The "additive within the range, with one exception" claim undersold it.** Only `hash` and `verify` span the whole range; the other twelve entries begin at named releases. Two further changes are type-only with unchanged arity — `hash_with_salt`'s salt `&[u8]` → `[u8; 16]` at 0.13.0, and `hash`/`verify`'s password `&str` → `P: AsRef<[u8]>` at 0.2.2 — which is why `parameter_types` is omitted rather than declared as one era's and wrong for the other. The header now says all of this.

### Unchanged and re-verified

`go test -race ./...` 31 ok, 0 failures, no race. `make lint` at the pinned golangci-lint: 0 issues. All 14 keys still resolve at their declared arities and nowhere else.

On `version_range` spanning releases where most entries do not exist: I have left it as `>=0.1.0,<0.20.0` and documented what actually begins where. A contract for a spelling that does not yet exist matches nothing, the field is parsed but never read for matching, and the sibling KBs do the same — so narrowing it would buy accuracy in a field nothing consumes while losing the record of which releases the family covers.

agustingroh:
Third review round. Two substantive defects, both real, both fixed — and the second one meant this PR was not delivering the thing its first sentence promises.

### The "exactly 14 entries" guard did not count the KB

Round two's comment claimed it "counts what the KB actually holds". It did not: it swept a hardcoded list of 21 names across arities 0–6. Anything outside that list or that range walked straight past it. Confirmed — all four of these left the suite green:

- `bcrypt.hash_str` / 2 (name not listed)
- `bcrypt.b64_encode` / 1 (name not listed)
- `bcrypt.hash` / 7 with `return: {type: bool}`, `role: config` (arity outside the sweep)
- `bcrypt.bcrypt` / 9 with `return: {type: none2}` (both)

The last two are round one's invented-return-token defect re-entering through the hole in the guard meant to close it. `countBcryptContracts` now walks `kb.Contracts` directly. Re-verified: the unlisted-name and out-of-range-arity entries both fail the suite now.

### The KB declared no signatures at all — which is what the PR is for

`bcrypt.yaml` omitted `parameter_types` and `canonical_return_type` on all fourteen entries, and justified it with a claim I had checked badly:

> an external call renders as `name(?, ?)` in that export whether or not a contract exists — a resolved `scrypt` call renders identically

`scrypt`'s contract does not declare them either, so both rendered `(?, ?)` and I read that as "the export always does this". The repo says otherwise in the sibling test this file was modelled on, and `internal/scan/export.go` confirms the mechanism: `canonicalMatches[0].ParameterTypes` and `.CanonicalReturnType` feed `canonicalSignature()`, which substitutes `?` only for empty slots.

Eleven of the fourteen entries have one unambiguous signature across the range they are reachable in, and now declare both fields. Before and after, same probe consumer:

```
bcrypt.hash_with_salt(?, ?, ?)                    →  (unchanged — see below)
bcrypt.non_truncating_hash(?, ?)                  →  bcrypt.non_truncating_hash(impl AsRef<[u8]>, u32): bcrypt::BcryptResult<String>
bcrypt.non_truncating_verify(?, ?)                →  bcrypt.non_truncating_verify(impl AsRef<[u8]>, &str): bcrypt::BcryptResult<bool>
bcrypt.bcrypt(?, ?, ?)                            →  bcrypt.bcrypt(u32, [u8; 16], &[u8]): [u8; 24]
```

The three still omitting them are the three where method plus arity does not identify one signature: `hash` and `verify` (password `&str` → `P: AsRef<[u8]>` at 0.2.2) and `hash_with_salt` (salt `&[u8]` → `[u8; 16]` at 0.13.0). Declaring either era's types there would be right for half the range and wrong for the other. The header now says that instead of the false rendering claim, and `impl AsRef<[u8]>` follows the spelling already used elsewhere in the KB.

### Two exported fields nothing pinned

`ParameterContract.Name` and `Return.Confidence` both reach the exported JSON and neither was asserted, so `name: hash` → `name: ciphertext` and `confidence: high` → `low` were silent. Both are now in the tables.

Writing that assertion surfaced something worth keeping: **the argument name and the contributed property are not the same thing**. `verify`'s second argument is named `hash` in the crate and contributes a `passwordHash`. My first version equated them and failed correctly. The two are now pinned separately, with the distinction spelled out.

### Body claims synced

Three statements the earlier rounds fixed in the header but not in the body: the raw primitive dated to 0.2.0 (it is 0.5.0), "additive apart from that one arity change" (only `hash` and `verify` span the range), and "whole surface is free functions" (the `HashParts` accessors and `Version` are public and skipped). Also removed the now-false "asserts exactly 14 entries" phrasing, which is true again only because the counter was rewritten.

`go test -race ./...` 31 ok, no race. `make lint` 0 issues.

PR #416, opened 2026-09-03
feat(callgraph): add aws-sdk-kms contract knowledge base

agustingroh:
Adds a callgraph contract KB for `aws-sdk-kms` — the generated AWS SDK for Rust's client for AWS
Key Management Service — so a consumer's encrypt, decrypt, re-encrypt, sign, verify, MAC, data-key,
shared-secret and random call sites resolve to a declared signature instead of an empty one.
92 entries.

The cryptography is the service's, not the caller's: every operation is an HTTPS request to KMS and
the key material never reaches the process, so nothing here claims a local runtime provider.

## Why the 68 setters are declared, which is the point of the file

Every KMS call is a fluent chain, so the operation resolves and everything after it falls back to
the **scanned crate's own name** — a *wrong* identity rather than a missing one, because it looks
resolved and matches no contract. Measured on a realistic consumer, through the full builder:

| State | Call sites resolved |
| --- | --- |
| before this contract existed | **6 of 21** |
| with the setters declared | **21 of 21** |

The fifteen that were lost came out as `<scanned crate>.key_id`, `<scanned crate>.plaintext`,
`<scanned crate>.signing_algorithm` and so on. The setters carry `role: config`: they are not
operations and are not reported as any, they are the links a chain travels to reach one.

## Why the floor is 0.23.0, and it is a compromise stated rather than hidden

| Version | Why |
| --- | --- |
| `0.0.0` | an **explicit placeholder** — its whole source is a `lib.rs` whose doc comment reads "This package is a placeholder for the future AWS SDK for Rust supporting KMS" |
| `0.2.0` | the alpha-era client exists, but `generate_mac`/`verify_mac` do not, and `client.encrypt()` returns `fluent_builders::Encrypt<C, M, R>` — a generic with three type parameters, not the type this file names |
| `0.23.0` | `generate_mac` and `verify_mac` arrive; the first version in which every operation named here exists |
| `1.1.0` | the first 1.x — there is no 1.0.0, it was never published — and the builders are renamed to `operation::<op>::builders::<Op>FluentBuilder`, the spelling declared here |

So the **identities** resolve from 0.2.0 but the **return types** are only correct from 1.1.0. A
consumer on 0.23.0–0.x gets the right key and a return type naming the wrong builder. A cleaner
split is inexpressible: `yamlContract` carries no per-entry version field, and
`validateLibraryUniqueness` rejects two files declaring the same library name.
`derive_shared_secret` is newer still — absent at 1.30.0, present at 1.35.0.

## Three tests, each checked against a mutation

They run through `BuildFromDirectories`, not the parser alone, because
`rust_fluent_chain_qualifier_test.go` records that a post-build pass can undo what the parser got
right.

1. **Every identity the builder emits joins exactly one contract** with the expected role and
   library — and every `aws-sdk-kms` key in the KB must be either exercised by the test source or
   explicitly listed as reachable-but-unexercised, so an added contract cannot pass unnoticed.
2. **No call site in a realistic consumer may fall back to the scanned crate.** This is the
   assertion that would have caught the state this KB started in, where 15 of 21 sites came out
   owned by the consumer's own package.
3. **Every declared arity is asserted against the crate's real signatures.** This one exists
   because arity is *documentation* for Rust: callees carry no encoded arity, every lookup arrives
   with `-1`, and `ContractsFor` resolves by name — so a wrong arity changes no match and is
   invisible to tests 1 and 2.

All four mutations a review used to show a sibling family's tests were inert — delete an entry,
flip a role, change an arity, append a contract for a module that does not exist — fail here.
Verified by applying them.

## Verification

```
$ GOPRIVATE='github.com/scanoss/*' go test -count=1 ./...            -> 31 ok, 0 FAIL
$ ./bin/golangci-lint run   (clean GOLANGCI_LINT_CACHE + GOCACHE)    -> 0 issues.
```


agustingroh:
## Review response

Ten findings, nine of them real, and one of them behavioral rather than cosmetic — which is the
part I had argued away. All fixed in `409b0ee`; the KB is now 171 entries.

### 1. Six wrong arities, and arity is not documentation-only after all

Confirmed. All six `*encryption_context` setters are `(mut self, k: impl Into<String>, v: impl
Into<String>)` — arity **2**:

```
encrypt/builders.rs:233        decrypt/builders.rs:159
re_encrypt/builders.rs:167     re_encrypt/builders.rs:314
generate_data_key/builders.rs:215   generate_data_key_pair/builders.rs:145
```

Identical from 0.23.0 to 1.111.0, so not version drift. Fixed to 2.

The part I had wrong is the reasoning, not just the number: I wrote that arity is documentation for
Rust because the call-graph path arrives with `-1` and resolves by name. You showed the **export**
path passes the real argument count and `rustContractsFor` bails before the name fallback when
`arity >= 0`, so all six matched nothing. The claim that a wrong arity is harmless was the
load-bearing error; it is now stated the other way round in the file.

### 2. The arity test hardcoded the wrong value and blocked the fix

Confirmed, and it is the sharpest process finding here: applying the *correct* value made the test
fail. It was a second hand-typed copy of the YAML claiming to come from the crate. It is gone.

### 3. Role-flipping passed on 78% of the KB, and the `unexercised` integers were dead data

Both confirmed. There is now **one** table asserting arity, role, `return.type` and
`canonical_return_type` for **all 171 entries**, and the loop checks every one. I re-ran your
mutation set plus the ones you suggested trying:

```
flip role on an UNEXERCISED entry (verify.key_id)   CAUGHT
bogus canonical_return_type (verify.send)           CAUGHT
bogus return.type on an UNEXERCISED entry           CAUGHT
revert an arity to the wrong value (2 -> 1)         CAUGHT
correct state                                       green
```

### 4. Roles contradicted the Go KMS KBs merged one PR earlier

Confirmed and aligned: `generate_data_key` and `generate_data_key_pair` are `factory`,
`generate_random` is `output`, matching `go/aws-kms-v2.yaml` and `go/aws-kms.yaml`. Your point about
`Contract.Role` flowing to the served surface as `Category` is what makes it matter — the identical
KMS call would have got a different category purely by consumer language. I had not looked at the
sibling KB at all.

### 5. The builder rename landed at 0.25.0, not 1.1.0

Confirmed by bisection, and this one embarrasses the whole argument I built on it:

```
0.24.0  pub fn encrypt(&self) -> fluent_builders::Encrypt
0.25.0  pub fn encrypt(&self) -> crate::operation::encrypt::builders::EncryptFluentBuilder
```

So **two** published versions carry the old spelling, not nineteen, and I had presented a
nineteen-release "stated compromise" that does not exist. Your further point is also right and
worse: at 0.23.0/0.24.0 none of the setters would join, so the chain collapses to exactly the state
the KB exists to fix — not merely "a return type naming the wrong builder". The floor is **0.25.0**
and there is no compromise to state.

### 6. `canonical_return_type` and 72 `return.type` values unguarded

Confirmed and covered by the single table in finding 3.

### 7. No `parameters:`/`contributes:`, against a precedent I set myself

Confirmed, and you quoted my own comment back at me from the neighbouring KB, which is the right
way to make the point. Added: the four `*_algorithm` setters and `key_spec`/`key_pair_spec`
contribute `algorithm` by `argument_type`, and `number_of_bytes` contributes `keySize` by
`argument_byte_length` — the case that field is documented for. 11 `parameters:` blocks.

### 8. The headline was off by one and mixed two definitions

Confirmed. It is **7 → 21** carrying an aws-sdk-kms identity, or 6 → 20 joining a contract; `(6, 21)`
is not self-consistent. The seventh was `aws_sdk_kms::primitives.Blob.new`, which resolves without
any contract and is in no KB. Corrected in the file, the commit and the body, with the reason.

### 9. 70 `set_*` setters and 11 `customize()` undeclared

Confirmed, and your proof is what settled it: an undeclared link collapses the rest of the chain, so
a consumer using the documented `set_key_id(Some(k))` override lost `send()` entirely. All of them
are declared now — that is the jump from 92 to 171 entries — because they return `Self` exactly as
their plain counterparts do. Declaring them is functional, which is also the answer to whether
`dry_run` and friends belong here: you reached the same conclusion independently.

### 10. Internal jargon in a public artifact

Fair. "a sibling family" is the same class of thing as "the matrix" and is gone; the comments now
say "another Rust KB", which is meaningful without the campaign context.

### Verification

```
$ GOPRIVATE='github.com/scanoss/*' go test -count=1 ./...                        -> 31 ok, 0 FAIL
$ ./bin/golangci-lint run   (clean GOLANGCI_LINT_CACHE + GOCACHE)                -> 0 issues.
```

Rebased onto current `main`, which resolved the CHANGELOG conflict you flagged and also picked up
`fix(callgraph): take a Rust associated function's return type from its contract` — the fix for the
issue this family's sibling filed. All 31 packages still pass with it, so the measurements above
hold on the new resolver.

Your "clean after real effort" list is the part I found most useful to read: the key-joining
analysis, the two inexpressibility claims, and the confirmation that `version_range` is never
consulted at match time all told me which of my arguments were load-bearing and which were
decoration.

Issue #478, opened 2026-09-14
Node/JavaScript callgraph parser: identity collides across files, and methods resolve only in class declarations


agustingroh:
## Why this comes before any Node family

Node is where Rust and Python were before their parser passes, and the gap is not
where the line counts suggest.

| | Node | Python | Rust | Java |
|---|---|---|---|---|
| parser lines | **650** | 3125 | 2379 | 4009 |
| resolution specs | **none** | 3 | 3 | — |
| contract files | **1, empty** | 38 | 152 | 38 |

`internal/callgraph/contracts/node/` holds a single `bootstrap.yaml` with
`contracts: []`, whose own comment reads *"remove this file when the first Node
library KB lands."* No Node library has ever been contracted.

## What was measured, and how

The authoritative node-type inventory was extracted from the `parser.c` that
**crypto-finder itself compiles** — not from an upstream repo — so it is exactly
the grammar in play: **170 named node types for JavaScript, 268 for TypeScript.**
`node_parser.go` names 28 of them.

That static list is a hypothesis, not a result, so 23 single-construct probes
were scanned with `--export-callgraph`, each carrying the same
`crypto.createHash('sha256')` call so that only the construct varies. Two layers
were read separately: whether a rule matched, and whether the parser bound the
call to a containing function.

**A first run of this said 2 of 23 resolved. That number was wrong and is
retracted.** Every probe named its function `f`, and the identity key is scoped
to the *directory*, so all 23 collided on `nodegrammar/src.f` and one survived.
The contradiction that exposed it: byte-identical code resolved in one run and
not in another. With unique names the real figure is **19 of 23**.

## Defect 1 — function identity collides across files in one directory

Two files in the same directory declaring a same-named function produce **one**
function. Minimal reproduction:

    src/alpha.js   export function hash(d){ return crypto.createHash('sha256')... }
    src/beta.js    export function hash(d){ return crypto.createHash('md5')... }

Both files match at the rule layer. The export contains exactly one function,
`nodecollide/src.hash`, pointing at `beta.js`; `alpha.js`'s finding comes back
`unresolved_reason: no_containing_function`.

This is routine Node layout — `index.js` and `utils.js` each with a `hash()` —
and it silently drops one of them. **Highest severity of the four: it is not a
missing construct, it is a wrong answer.**

## Defect 2 — methods are recognised only inside `class_declaration`

Three idiomatic shapes bind no containing function, all one cause:

| shape | rule layer | parser |
|---|---|---|
[...]

PR #481, opened 2026-09-14
feat(dependency): resolve node/npm dependencies for --scan-dependencies

### LINKED ISSUE #480: Node/npm dependency scanning is unimplemented, and --scan-dependencies reports success without it
Tier 0 requires local CLI dependency scanning for five ecosystems. Four have a resolver. Node has none, and the gap is invisible from the outside.

## Measured

`internal/dependency/` holds six `Resolver` implementations, and `internal/cli/scan.go:969-972` registers four ecosystems:

| ecosystem | resolver | how it obtains the tree |
|---|---|---|
| go | `GoResolver` | runs `go` |
| rust | `CargoResolver` | runs `cargo metadata --format-version=1` |
| java | `JavaResolver` -> `GradleResolver` / `MavenResolver` | delegates by build system |
| python | `PipResolver` | inspects the environment |
| **node** | **none** | — |

No file in `internal/dependency/` mentions `package.json`, `package-lock.json` or `node_modules`.

## The failure is silent, which is the actual defect

`--scan-dependencies` on a Node project:

```
WRN No resolver for ecosystem, skipping dependency scan
    error="no resolver registered for ecosystem \"node\"" ecosystem=node
 • Files with findings: 0
 • Total crypto assets: 0
```

Exit code 0. A Node project with a full cryptographic dependency tree is indistinguishable from one with no cryptography at all. `scan.go:975` logs a warning and continues by design, which is right for an ecosystem nobody claims to support — but Node IS claimed:

> For Java, Python, Rust, Go, **and Node.js/TypeScript**, prove local CLI dependency scanning. C/C++ dependency scanning is explicitly deferred.

That note appears in every Tier 0 node issue. C/C++ is deferred explicitly; Node is required and absent.

## Why it blocks the whole node queue, not one family

Without a resolver, `Dependency.Dir` is never produced for a Node dependency, so the finder never parses the dependency source. A consumer writing `forge.md.sha256.create()` gets the call matched at the rule layer and then nothing behind it: no reachability into the library that implements the primitive. That is the evidence Tier 0 asks each family to produce, so the acceptance criterion "Local CLI dependency scanning is covered where applicable" cannot be met for **any** of the 63 open `node` + `ready-for-agent` issues.

## Scope

Add `NpmResolver` implementing `dependency.Resolver` and register it for `node`:

- root coordinate and version from `package.json`
- dependency tree and versions from `package-lock.json` (lockfile v2/v3 `packages`, keyed by install path, and the v1 nested `dependencies` shape)
- `Dir` from the install path, so nested and deduped installs resolve to the directory actually on disk rather than a guessed `node_modules/<name>`
- `Graph` / `VersionedGraph` from the lockfile

npm is the simplest of the five because the sources are already on disk: there is no download step.

**The one real design question is what happens when `node_modules` is absent.** Resolving from a lockfile alone yields dependencies with no source directory, which is a tree that looks resolved and parses to nothing — the same silent-success shape this issue is about. The resolver should make that state explicit rather than return a hollow result.

Out of scope: yarn and pnpm lockfiles, workspaces, and any network fetch.

## Acceptance criteria

- [ ] `NpmResolver` resolves a project with `package.json` + `package-lock.json` + installed `node_modules` to dependencies whose `Dir` exists on disk, with versions preserved.
- [ ] Both lockfile shapes are covered: v2/v3 `packages` and v1 nested `dependencies`.
- [ ] A nested or deduped install resolves to the directory that actually holds the source.
- [ ] Absent `node_modules` produces an explicit, named outcome rather than a dependency list with empty directories, and a test pins it.
- [ ] `Graph` and `VersionedGraph` carry parent -> child edges with versions.
- [ ] The resolver is registered for `node` in `scan.go`, and an end-to-end `--scan-dependencies` run on a Node fixture produces dependency evidence instead of the skip warning.
- [ ] Tests assert against a real resolved tree; a test that would pass against an empty result is not sufficient.

agustingroh:
Closes #480.

`internal/dependency/` shipped resolvers for go, rust, java and python. A node target fell through to a warning and **exit 0**:

```
WRN No resolver for ecosystem, skipping dependency scan
    error="no resolver registered for ecosystem \"node\"" ecosystem=node
```

A Node project with a full cryptographic dependency tree was indistinguishable from one with no cryptography, and nothing in the output said the scan had not happened. Tier 0 requires local dependency scanning for Java, Python, Rust, Go **and Node**; C/C++ is the only ecosystem deferred explicitly, so this blocked the acceptance criterion on all 63 open node issues.

## What it does

`NpmResolver` reads `package.json` + `package-lock.json` and maps every production dependency to the directory npm installed it in. It runs no tool, unlike cargo/go/java: npm puts sources on disk at install time, so resolution is reading the lockfile and pointing at what is already there.

**Both lockfile shapes.** A v2/v3 `packages` map is keyed by install path and answers directly what the v1 nested `dependencies` shape needs a tree walk to answer, so v1 is flattened into the same path-keyed form and one code path serves both.

**Edges follow node's own resolution order** — the depender's own `node_modules` first, then each ancestor's, ending at the root. That walk is the non-obvious part: guessing `node_modules/<name>` reports the hoisted copy for both a nested and a hoisted install, and the two are different source at different versions.

**Dev dependencies are excluded** — build tooling that never ships, and walking it multiplies the scanned tree. Pinned by a test so it stays deliberate rather than incidental.

**No install fails loudly.** `dependency_scanner.go:202` turns a `Resolve` error into `CodeDependencyResolutionFailed`, so the cause surfaces. Returning dependencies with empty `Dir` fields would be a tree that looks resolved and parses to nothing — the same silent success this change exists to remove.

## Measured end to end, against a real `npm install` of node-forge

| project | main | this branch |
|---|---|---|
| with `node_modules` | exit 0, `No resolver`, never scans | exit 0, `depsScanned=1 depsSkippedNoSource=0` |
| without `node_modules` | exit 0, skipped silently | **exit 1**, `npm resolver: package-lock.json lists 1 dependencies but none are installed under node_modules in .: run ``npm install`` before scanning` |

Confirmed on a cold cache (`--findings-cache none`), since a cache hit can hide a scan that did not run.

## Tests

Eight, written failing first, each asserting against a real tree on disk rather than a parsed structure — `Dir` is the field that carries the evidence, so every assertion `os.Stat`s the file it claims is there.

| test | what it pins |
|---|---|
| `ResolvesInstalledTreeFromLockfileV3` | deps, versions, `Dir` holding real source, `Graph` and `VersionedGraph` edges |
| `ResolvesNestedInstallToTheDirectoryOnDisk` | two versions of one package resolve to two different directories; the edge points at the nested copy |
| `ResolvesLockfileV1NestedShape` | the v1 shape flattens to the same result |
| `AbsentNodeModulesIsAnExplicitFailure` | a lockfile with no install errors, and the message names `node_modules` |
| `MissingLockfileIsNamed` | a different cause says so |
| `DevDependenciesAreExcluded` | dev out, production in |
| `ScopedPackageResolves` | `node_modules/@scope/name` is two path segments |
| `Ecosystem` | registered name |

Each fix was mutation-tested — revert it and the test dies:

| mutation | result |
|---|---|
| guess `node_modules/<name>` instead of walking ancestors | **nested test fails** |
| stop excluding dev | **dev test fails** |
| return a hollow result instead of erroring | **absent-install test fails** |
| stop checking the directory exists on disk | **absent-install test fails** |

`go test ./internal/dependency/... ./internal/cli/...` green. `make lint` 0 issues.

Out of scope, deliberately: yarn and pnpm lockfiles, workspaces, and any network fetch.


agustingroh:
Follow-up commit: workspace members were resolved as external dependencies.

Found while re-checking what "workspaces are out of scope" actually did. It did not decline to handle them — it returned a wrong answer. On a two-package monorepo:

```
DEP module="inner" version="1.0.0" dir=/packages/inner
DEP module="inner" version=""      dir=/node_modules/inner
WORKSPACE_MEMBERS=0
```

npm keys a workspace member twice in a v3 lockfile: once at its real path, and once as a `"link": true` entry under `node_modules` that points at it and carries no version. Both were taken, so the consumer's OWN source was reported under an external coordinate, twice, and one copy had no version to form a coordinate from at all. For a campaign that distinguishes author code from dependency code, that is the wrong side of the line.

Link entries are now skipped — their target is already keyed by its own path — and a lockfile path outside `node_modules` becomes a `WorkspaceMember`, which is the same distinction `CargoResolver` draws between a workspace crate and a dependency.

Pinned by `TestNpmResolver_WorkspaceMembersAreNotDependencies`, which also asserts no resolved dependency has an empty version. Both halves mutation-tested: stop skipping `link` entries, or stop separating paths outside `node_modules`, and the test dies.

Yarn and pnpm lockfiles and network fetch remain out of scope, and now actually are.

agustingroh:
Review fixes pushed. Nine of the ten defects are addressed here; the tenth is split out as #482 because it cannot be fixed inside a resolver.

**The empty PURL (HIGH).** `purl.Dependency` had no npm case, so every npm dependency finding was exported with no coordinate, and rule-PURL version enrichment could never match because it compares `purl.Identity` of the rule against `purl.Identity` of the dependency and `""` is always skipped. Proven end to end on a consumer whose dependency matches a rule:

```
before:  {"module": "hashlib-probe", "version": "2.1.0"}
after:   {"module": "hashlib-probe", "version": "2.1.0", "purl": "pkg:npm/hashlib-probe@2.1.0"}
```

A scoped package keeps the `@` in its namespace (`pkg:npm/%40noble/hashes@1.3.0`), the form the mining service source resolver reads back. The test that asserted node produced an empty string is replaced — it pinned the defect.

**Dev dependencies are resolved (HIGH).** My consistency claim was wrong and the review measured it: Cargo runs bare `cargo metadata` and appends every package, Go runs `go list -m -json all`, pip lists the whole environment. Only the two Java resolvers narrow to compile scope. I had sided with the two that exclude against the three that include, in a crypto inventory tool, silently.

**Absent dependencies are classified, not counted (MEDIUM-HIGH).** The old warning fired on every healthy project — `npm install esbuild` leaves 25 of 26 `@esbuild/*` platform packages absent because npm gates them on os and cpu. A warning that always fires is no warning, and the silent success this PR exists to remove survived it: 48 of 50 required packages missing resolved 2 and exited 0. Optional and dev absences are expected now; an absent required package fails and is named.

**Edges.** peer and optional edges are included; the graph is built only from packages that resolved on disk, so no more phantom nodes; an edge names its target by the package it resolved to rather than by the depender's map key, which for an aliased install is the alias — five appeared in a stock `npm install eslint jest webpack` without anyone asking for them. Workspace members are graph parents now, so their dependencies are no longer orphans.

**npm 5.0 `"requires": true`** no longer aborts the scan.

**Tests.** Every defect above is pinned, plus the v2 lockfile shape the doc comment claimed and nothing covered. Worth stating plainly: nine of the ten were invisible to my original suite because I built its fixtures by hand, so they contained no alias, no optional platform package and no peer edge. The review found them by running against real trees and diffing against `npm ls --all --json`.

**#482** is the workspace-root one. The obvious fix — appending the target alongside the members — double-parses every member, because `analyzeDir` walks into subdirectories and npm keeps members under the root. With module-scoped identity that gives two identities for one function, which is the collision class #478 just removed. It needs a per-package exclusion in `callgraph.Builder`, so it is its own change.

`go test ./internal/dependency/... ./pkg/purl/... ./internal/cli/...` green, `make lint` 0 issues.

agustingroh:
Tenth defect folded in, and #482 closed rather than left open.

The reason is that **this PR is what introduces it**. Before it there was no node resolver, so `WorkspaceMembers` was never populated for a node target and the scan target was always the package root. Populating members is exactly what takes the root out of the graph. That makes it a defect of the feature being added here, not a pre-existing one I happened to find, so splitting it out was the wrong call.

`collectPackageSets` chose the members OR the scan target, never both — right for a Cargo virtual manifest, wrong for an npm workspace root that carries its own `index.js` and its own dependencies.

The fix is additive: `callgraph.PackageDir` gains `ExcludeDirs`, and the root is walked with the member directories excluded. Anything that leaves `ExcludeDirs` empty behaves exactly as before, so Cargo, Maven, Gradle and pip are untouched. The exclusion is necessary because `analyzeDir` walks pre-order into subdirectories and npm keeps members UNDER the root: `node_modules` is already excluded by default, member directories are not, so without it each member is parsed twice under a second import path and one function acquires two identities — the collision class #478 just removed.

The test carries its own positive control: it first asserts that WITHOUT the exclusion the member's one function holds TWO identities, so the exclusion assertion cannot pass for the wrong reason, then asserts identity uniqueness across the whole graph.

Two existing tests pinned the old behaviour and are updated, both named in the commit.

Full suite green across `callgraph`, `engine`, `dependency`, `scan`, `purl` and `cli`; `make lint` 0 issues.

PR #490, opened 2026-09-15
fix(skip): scan a package whose compiled output is its only source

### LINKED ISSUE #487: An npm package that publishes only dist/ or build/ mines to an empty graph, and the zero looks clean
Found while mining `npm:asn1js` for Tier 0 (scanoss/crypto-mining-service#248). asn1js itself performs no cryptography, so nothing was lost there — but the mechanism is general and the next package it hits may be a cryptographic one.

## Measured

`internal/skip/source_defaults.go:41-42` excludes `dist` and `build` from the source walk. That is right for a repository checkout, where they are generated output beside the sources.

**An npm registry tarball is not a checkout.** npm publishes what `files`/`.npmignore` allows, and for a large share of packages that is *only* the compiled output. asn1js is the clean case:

| version | directories in the tarball | functions in the mined graph | graph fragment |
|---|---|---|---|
| 2.4.0 | `src/`, `build/` | **235** | 22745 bytes |
| 3.0.0 | `build/` | **0** | 112 bytes |
| 3.0.5 | `build/` | **0** | 111 bytes |
| 3.0.10 | `build/` | **0** | 113 bytes |

The 2.x line ships `src/` and parses; the 3.x line ships only `build/` and parses nothing.

Direct positive control, two identical files:

```
/tmp/bx/build/a.js   crypto.createHash("sha256")...
/tmp/bx/src/a.js     crypto.createHash("sha256")...
```

```
default:                   Files with findings: 1   ->  src/a.js
--no-default-exclusions:   Files with findings: 2   ->  build/a.js, src/a.js
```

A crypto call inside `build/` is invisible by default.

## Why it matters, and why it is not urgent

The failure mode is the one this campaign keeps paying for: **the result is an empty graph and zero findings, which is indistinguishable from a package that genuinely uses no cryptography.** Nothing in the output says the tree was skipped.

Checked against the four npm families mined so far — node-forge ships `lib/`, and ethereumjs-util, @ethereumjs/util, @libp2p/crypto and libp2p-crypto all ship `src/` alongside `dist/`, so their sources were read and their zeros are genuine. asn1js 3.x is the only one affected, and it has no cryptography to lose. So this is a latent hazard, not a live miscount.

## What would fix it

The exclusion is about *provenance*, not directory name: generated output beside sources is noise, generated output that is the ONLY thing present is the package. Some options, in rough order of how much they change:

- When a scan target contains no source outside the excluded directories, do not treat them as excluded — and say so in the output.
- Let the mining path pass a different exclusion set than a repository scan, since it knows it is looking at a published artifact.
- At minimum, **report it**: a graph fragment with zero functions where files existed but were all excluded should be distinguishable from one where no files existed at all. Today both are the same empty fragment.

The last one alone would turn a silent zero into a visible one, which is the part that matters.

agustingroh:
Closes #487.

## What

`dist` and `build` are in `DefaultSkippedDirs` because in a **checkout** they hold generated copies of the sources beside them. In a **published npm tarball** they are frequently everything there is — the modern default for a TypeScript package — so the exclusion left the scanner with nothing to read.

## Measured

`@azure/cosmos` 4.10.1 ships only `dist/` (plus `package.json`, `README.md`, `LICENSE`). Same scan, same rules:

| exclusions | files | assets |
|---|---|---|
| default (before) | 0 | 0 |
| default (after) | 6 | 6 |
| `--no-default-exclusions` (control) | 6 | 6 |

The danger is not the missing six. It is that a zero meaning *"nobody read this tree"* is indistinguishable in a report from one meaning *"no cryptography here"*, and the second is what reaches a coverage tracker.

## How

`skip.BuiltOutputOnlySource(targetDir)` walks the target once and answers one question: **is there any source outside the built-output directories?**

- Yes → defaults unchanged. One source file anywhere outside, including under `src/`, is enough.
- No, and a built-output directory holds source → that name alone is dropped from the exclusions.

Conservative by construction:

- Only `dist` and `build` are eligible. `node_modules` and `vendor` hold *other packages*; `docs` holds prose.
- Only a **top-level** `dist` counts. `packages/x/dist` beside `packages/x/src` stays excluded.
- `package.json`, `README`, `LICENSE` are **not** source — otherwise the tarball layout would never be rescued.
- Every other default (`node_modules`, `vendor`, `build` when it is not the artifact, the generated-stub globs) is untouched.
- `--no-default-exclusions` is unaffected.
- The walk short-circuits on the first source file found outside, and is bounded at 50k entries.

The scan logs the rescue and names the `--exclude` that overrides it.

## Tests

`internal/skip/built_output_test.go` (9 table cases) and `internal/cli/scan_flags_test.go` (the CLI seam). Written failing-first; mutation-checked at both layers — stubbing `BuiltOutputOnlySource` to return nil fails 4 skip cases and the CLI rescue case.

Full suite green, `make lint` 0 issues.


agustingroh:
Review pass applied. Six defects found, six fixed, each with an end-to-end check rather than a unit test alone.

**1. `crypto-finder scan ./pkg` silently defeated the fix (High).** `WalkDir` joins — and so *cleans* — every child path, so trimming the raw target spelling as a prefix matched nothing for `./pkg`, `pkg/.`, `pkg//`, `/abs/./pkg`. Every file then read as top level and no rescue fired. Now `filepath.Clean` on the root plus `filepath.Rel` per entry. Measured on one `dist`-only fixture, changing only the spelling: all six spellings now report 6 findings (previously 4 of the 6 reported 0).

**2. Source the scan will never read suppressed the rescue (High).** The pre-pass counted any file outside `dist`/`build` as source without applying the exclusion list it was deciding about. Installing dependencies cancelled the fix — which the npm resolver in this same changelog requires. `node_modules`, `vendor`, the other `DefaultSkippedDirs` and the default test directories are now skipped by the pre-pass, and `*.d.ts` is not source (a declaration file holds no executable code, so `dist/` + root `index.d.ts` — an ordinary published layout — would have left the scan reading types and reporting the same ambiguous zero). Measured: `+ node_modules` 6, `+ index.d.ts` 6, `+ test/t.js` 6; all three were 0.

**3. One unreadable directory turned a checkout into a rescue (Medium-High).** `fs.SkipAll` makes `WalkDir` return nil, so the "unreadable tree" guard was dead and the partial state was used as a complete answer. The scan now records that it was cut short and returns nil — the defaults. Same for **4**, the 50k entry cap, which had the same shape: `dist`/`build` sort before `src`/`lib`/`packages`, so the ordering worked against the check.

**5. A file target inside `dist` still reported a silent zero (Medium).** The CLI maps a file target to its parent, so `scan pkg/dist/index.js` asked about `pkg/dist`, where everything is top level. A target named at the built output *is* the artifact and is now read: 1 finding, was 0.

**6. The rescue notice was invisible (Low-Medium).** `Info` → `Warn`, matching the neighbouring `--no-default-exclusions` notice. Dropping a built-in exclusion should be visible in the run that did it.

Also corrected the inaccurate "every other walk skips dot directories" comment (`internal/language/enry.go` has no such rule; the callgraph builder does).

**Tests.** Every branch the review found uncovered is now pinned: target spelling (6 forms), target-is-the-built-output plus the empty case, unreadable directory (skipped under root), the entry cap (passed in, not a mutated global), a nonexistent target, and `node_modules`/`.d.ts`/`test/`/`vendor`/dot-dir as table cases. `TestDefaultPatternsForTargetIsUnchangedForACheckout` now compares element by element — the `len != len` assertion would have passed an implementation that dropped one name and added another. `internal/skip` at 91.7%; the two remaining uncovered lines are the `WalkDir`/`filepath.Rel` guards, which the code comments now identify as guards rather than reachable error paths.

Regression check with a finding-bearing file, same bundle in both trees: `dist/` only → 1, `dist/` + `src/index.ts` → 0. Full suite green, `make lint` 0 issues, `-race` clean.

Not changed, and worth recording: `internal/callgraph/rust_manifest.go` and `internal/callgraph/python_dependency_type_resolver.go` still splice raw `skip.DefaultPatterns()`. Neither affects this case — the callgraph builder's matcher is always overridden via `SetSkipMatcher` — but the rescue is a property of one code path today. The `--scanner semgrep` adapter also passes no flag disabling semgrep's bundled `.semgrepignore`, unlike the opengrep adapter; if that list carries `dist/` the fix would be a no-op there. Unverified (semgrep is not installed here) and out of scope for this PR; opengrep is the default and is correct.

PR #492, opened 2026-09-15
feat(callgraph): add the @azure/keyvault-keys contract knowledge base

agustingroh:
Tier 0 family `npm-azure-keyvault-keys` (tracker: scanoss/crypto-mining-service#251). Contract phase, paired with scanoss/crypto_rules#371.

## What

**39 entries under `internal/callgraph/contracts/node/azure-keyvault-keys.yaml`** for `@azure/keyvault-keys` >= 4.0.0, read off the package's own `.d.ts` rather than the Azure documentation.

Azure Key Vault performs its cryptography **server-side**, so no entry names an algorithm: the consumer passes it at the call site and which implementation runs is a property of the vault. What this KB types is the consumer's *request*.

Arity is part of the key. Every constructor and method carries an optional trailing options argument, so each is declared at every arity a consumer can write. `encrypt`/`decrypt` take **either** a parameters object or the algorithm positionally — both overloads are declared in `cryptographyClient.d.ts` and both appear in published consumers — while `sign`, `verify`, `wrapKey` and `unwrapKey` take the algorithm positionally only.

`getCryptographyClient` is contracted although it performs no cryptography: it *returns* a `CryptographyClient`, so it is the third way a consumer obtains one. Key listing, deletion, recovery, backup/restore, rotation policies and property updates are deliberately **not** contracted — they administer the vault and perform no cryptographic operation.

## Measured: this KB does not join yet, and the reason is in the parser, not in the KB

Stated plainly rather than discovered later. A consumer written in the most favourable possible shape — `require` at the top, `new CryptographyClient(...)` directly, the call immediately on that variable:

```js
const { CryptographyClient } = require("@azure/keyvault-keys");
const cryptographyClient = new CryptographyClient(keyUrl, cred);
const signed = await cryptographyClient.sign("RS256", data);
```

exports this symbol, **with this KB present**:

```
direct/index.sign          <- const signed = await cryptographyClient.sign("RS256", data);
```

not `@azure/keyvault-keys.CryptographyClient.sign`. Positive control on the same binary, same flags, a `node-forge` consumer:

```
node-forge.pki.certificateFromPem   <- const cert = forge.pki.certificateFromPem(pem);
node-forge.md.sha256.create         <- const md = forge.md.sha256.create();
```

So the mechanism that normalises a call site to the library coordinate is **namespace-import resolution** (`forge.x.y()` where `forge` is the imported binding), and a contract's `return.type` is not propagated to a variable receiver in JavaScript. `node-forge` joins because consumers write namespace accesses; this library is used by constructing a client and calling methods on the instance, which is the receiver-resolution gap the `@aws-sdk/signature-v4` tracker entry already records as deliberately not implemented.

## Why it ships anyway

1. `12-api-contract-join.sh` now passes for this family: **41 rules, 9 distinct api values, 170 contract entries, every api value has an entry to join**. Without this file every rule's `api` synthesizes an entry point that can never join anything.
2. The entries are correct today and independent of the parser. The day receiver typing lands, this family joins with no further work.

`go test ./internal/callgraph/...` green.


agustingroh:
Validated as far as is possible locally, without mining or stitching. Three separate questions, because the first two are indistinguishable from a scan result — no finding carries the coordinate either way.

**1. Is the KB in the binary?** Yes. `LoadEmbedded("node")` returns 170 keys, 39 of them this family's, indexed with arity: `@azure/keyvault-keys.CryptographyClient.sign#2`, `…KeyClient.importKey#3`. Pinned by the test in this PR; removing the YAML fails 14 assertions.

**2. Are the entries right?** Validated in both directions.

*Contract → package.* Every contracted method and arity checked against the package's own published type declarations (`dist/commonjs/*.d.ts`, 4.10.2), counting each signature's required and optional parameters: **32 methods, 0 problems**. No arity is declared that the library does not accept.

*Real consumer code → contract.* Every call site this family's rules match across the 15-consumer M20 draw, with the arguments counted at the call site by balanced-paren splitting rather than by reading:

| shape real consumers write | occurrences | contract declares |
| --- | --- | --- |
| `unwrapKey(alg, key)` | 10 | 2, 3 |
| `wrapKey(alg, key)` | 10 | 2, 3 |
| `sign(alg, digest)` | 5 | 2, 3 |
| `verify(alg, digest, sig)` | 4 | 3, 4 |
| `encrypt(params, opts)` | 1 | 1, 2, 3 |
| `decrypt(params, opts)` | 1 | 1, 2, 3 |

**31 call sites, 6 distinct shapes, 0 without a contract entry.** So the day a variable receiver resolves, this KB joins real published code — not a fixture written to match it. Those six arities are now pinned in the test with their counts.

**3. Does it join today?** No, and the cause is the parser, not this file. Re-checked while validating the above: the constructor is not emitted with the coordinate either — the graph for a direct-construction consumer carries only `direct/index.sign` and `direct/index.wrapKey`, with no `@azure/keyvault-keys.CryptographyClient.<init>` anywhere. Positive control on the same binary: a `node-forge` consumer emits `node-forge.pki.certificateFromPem`. Namespace-import resolution is what normalises a call site to a coordinate today.

What cannot be validated locally: that a mined payload actually stitches through these entries. That needs the mining stage, which C-1 suspends for rules families, so it is stated as not run rather than reported as passed.

Issue #534, opened 2026-09-18
main is red: TestLoadEmbeddedNodeFastJwt sees jsonwebtoken keys after the jsonwebtoken KB landed


matiasdaloia:
## Summary

`main` is red. `TestLoadEmbeddedNodeFastJwt` fails on a clean checkout of `22a9533`, with no local changes and no external toolchain involved. The test loads the embedded Node knowledge bases and asserts every contract key belongs to `fast-jwt`; eight `jsonwebtoken.*` keys are in the merged set instead.

## Reproduce

Clean checkout at `22a9533` (`feat(contracts): add jsonwebtoken Node contract KB (#530)`), then:

```
go test ./internal/callgraph/contracts/ -run TestLoadEmbeddedNodeFastJwt
```

```
--- FAIL: TestLoadEmbeddedNodeFastJwt (0.01s)
    node_fast_jwt_kb_test.go:81: key "jsonwebtoken.sign#3" belongs to jsonwebtoken., a different package
    node_fast_jwt_kb_test.go:81: key "jsonwebtoken.decode#2" belongs to jsonwebtoken., a different package
    node_fast_jwt_kb_test.go:81: key "jsonwebtoken.verify#3" belongs to jsonwebtoken., a different package
    node_fast_jwt_kb_test.go:81: key "jsonwebtoken.sign#4" belongs to jsonwebtoken., a different package
    node_fast_jwt_kb_test.go:81: key "jsonwebtoken.sign#2" belongs to jsonwebtoken., a different package
    node_fast_jwt_kb_test.go:81: key "jsonwebtoken.decode#1" belongs to jsonwebtoken., a different package
    node_fast_jwt_kb_test.go:81: key "jsonwebtoken.verify#4" belongs to jsonwebtoken., a different package
    node_fast_jwt_kb_test.go:81: key "jsonwebtoken.verify#2" belongs to jsonwebtoken., a different package
FAIL	github.com/scanoss/crypto-finder/internal/callgraph/contracts	0.007s
```

No network, no `mvn`, no JDK, no opengrep. It is pure embedded-YAML loading, so it fails the same way in CI.

## What this looks like

[...]

Issue #391, opened 2026-09-01
A tree with no root build manifest fails the ENTIRE scan under --scan-dependencies, discarding completed detection


matiasdaloia:
## Summary

With `--scan-dependencies`, a Java source tree that has **no Maven/Gradle manifest at the scan root** aborts the whole scan with `java_build_tool_unknown` — **after detection has already completed** — and writes no output at all. The crypto that detection found is thrown away.

There is already a correct path for "nothing to resolve": `progress.Skip("dependencies", "scan", "ecosystem_unknown")` in `internal/cli/scan.go`. This failure just does not route to it.

## Measured

crypto-finder **0.25.3**. Target is a single file, `Use.java`, containing `Cipher.getInstance("DES/ECB/PKCS5Padding")`. No manifest anywhere.

```
crypto-finder scan /w/D --error-format json --progress --timeout 10m \
  --scan-dependencies -o /w/outD/i.json
```

Progress stream:

```json
{"phase":"scan","status":"started"}
{"phase":"detection","status":"started","parent_phase":"scan"}
{"phase":"rules","status":"completed","parent_phase":"scan","duration_ms":744}
{"phase":"detection","status":"completed","parent_phase":"scan","duration_ms":5463}
{"phase":"dependencies","status":"started","parent_phase":"scan"}
{"phase":"dependencies","status":"failed","parent_phase":"scan","duration_ms":0}
{"phase":"scan","status":"failed","duration_ms":5464}
```

```json
{"code":"java_build_tool_unknown","stage":"dependency","retryable":false,
 "message":"could not detect a supported Java build tool in /w/D",
 "raw_error":"dependency scan failed: could not detect a supported Java build tool in /w/D"}
```

Exit status 1. **No interim and no callgraph written.** Detection spent 5.4s and succeeded; all of it is lost.

## Blast radius

This is not an exotic shape. Every variant below fails identically:

| target | result |
|---|---|
| one `.java` at the root, no build file | `java_build_tool_unknown`, whole scan fails |
| `.java` under `services/payments/src/`, no manifest anywhere | same |
| `services/{a,b}/pom.xml`, no root pom (monorepo) | same |
| root aggregator `pom.xml` + `<modules>` | works (1747 dependency findings) |

[...]

PR #535, opened 2026-09-18
fix(dependency): resolve module roots below a manifest-less scan root

### LINKED ISSUE #533: Dependency resolution only looks at the scan root, so module manifests one level down are invisible to it — even though detection already walked into those modules
## Summary

Detection walks the whole tree. Dependency resolution looks **only at the directory it was handed**. So a monorepo whose module manifests sit one level down gets zero dependency coverage, even though detection has already scanned inside those very modules.

`DetectEcosystem` (`internal/scan/ecosystem.go`) is a series of `os.Stat` calls at the scan root, and `DetectJavaBuildTool` (`internal/dependency/java_resolver.go:80`) does the same. Neither ever looks below. A `services/ledger/pom.xml` is invisible to both.

## Measured

crypto-finder **0.25.5-deps**. Target is a two-file tree with **nothing at the root**:

```
./services/ledger/pom.xml                        (commons-lang3 3.12.0)
./services/ledger/src/main/java/app/Use.java     (Cipher.getInstance("DES/ECB/PKCS5Padding"))
```

**Without `--scan-dependencies`** — detection reaches into the module and succeeds:

```json
{"phase":"rules","status":"completed","duration_ms":1237}
{"phase":"detection","status":"completed","duration_ms":5520}
{"phase":"dependencies","status":"skipped","details":{"reason":"not_requested"}}
{"phase":"scan","status":"completed","duration_ms":5560}
```

The interim names the file and the algorithm:

```
"services/ledger/src/main/java/app/Use.java"
"DES-ECB-PKCS5Padding"
```

**With `--scan-dependencies`** — same tree, same detection, then:

```json
{"phase":"detection","status":"completed","duration_ms":5423}
{"phase":"dependencies","status":"started"}
{"phase":"dependencies","status":"failed","duration_ms":1}
{"phase":"scan","status":"failed","duration_ms":5425}
```

```json
{"code":"java_build_tool_unknown","stage":"dependency","retryable":false,
 "message":"could not detect a supported Java build tool in /w",
 "details":{"target_dir":"/w"}}
```

No interim written. The two facts sit one millisecond apart in the same run: detection reported a finding **inside** `services/ledger/`, and resolution reported no build tool **at** `/w`, with `services/ledger/pom.xml` on disk the whole time.

## Why this is the common shape, not an edge case

A workspace *declared by a root manifest* already works — `buildWorkspaceMembers` (`internal/dependency/maven_resolver.go:118`), Cargo workspace members, npm workspaces, Gradle `WorkspaceMembers`. Multi-module support is there once a resolver is selected.

What has no path at all is a repository with modules and **no root manifest**: a polyglot monorepo (`services/api/pom.xml` beside `web/package.json`), independent modules with no aggregator, or Java parked under `services/` or `backend/`.

The tell: to get a repository like that scanned today you have to add an aggregator manifest at the root purely so detection can find it. Restructuring the repository to suit the scanner is the wrong way round.

## What blocks it

`Resolver.Resolve(ctx, targetDir)` takes one directory, and `ResolveResult.RootModule` is a singular string (`internal/dependency/resolver.go:59`). Everything else in that struct merges naturally across roots: `Dependencies` is a flat list, `Graph` and `VersionedGraph` are maps keyed by module path.

## Proposed shape

Bounded discovery that feeds the resolvers already present, rather than a new resolver contract:

1. When the root carries no manifest, walk to a bounded depth under the existing scan skip rules and collect candidate `(ecosystem, dir)` roots. `node_modules` and vendored trees are full of manifests, so this has to reuse the scanner's own exclusions.
2. Resolve each root with the existing resolver, under a cap on how many roots one scan will resolve. Each resolution shells out to `mvn`/`npm` and pulls source jars, so an unbounded fan-out is not affordable.
3. Merge: concatenate `Dependencies`, union the two graph maps. For the singular `RootModule`, the discovered roots fit the existing `WorkspaceMembers` semantics as stated on the field itself, *"When non-empty, all members are treated as user code for call chain tracing"* — which is exactly right for a monorepo's own modules.

## Relationship to #391

Distinct, and #391 comes first. #391 is that this failure aborts the whole scan instead of routing to the existing `progress.Skip("dependencies","scan","ecosystem_unknown")`; the run above reproduces it as a side effect. Fixing #391 makes `--scan-dependencies` safe to pass on any tree. This issue is what makes it *useful* on a monorepo.

matiasdaloia:
Closes #533.

## What changes for a user

A repository whose build manifests sit below the scan root now gets dependency coverage under `--scan-dependencies`, instead of failing the entire scan and throwing away detection findings that were already computed.

The reproduction from the issue, a two-file tree with nothing at the top:

```
./services/ledger/pom.xml                        (commons-lang3 3.12.0)
./services/ledger/src/main/java/app/Use.java     (Cipher.getInstance("DES/ECB/PKCS5Padding"))
```

Before, exit 1 and no report written at all:

```json
{"phase":"detection","status":"completed","duration_ms":2289}
{"phase":"dependencies","status":"started"}
{"phase":"dependencies","status":"failed","duration_ms":0}
{"phase":"scan","status":"failed","duration_ms":2290}
{"code":"java_build_tool_unknown","stage":"dependency","retryable":false,
 "message":"could not detect a supported Java build tool in /w"}
```

After, on the same tree with the production ruleset, exit 0:

```json
{"phase":"dependencies","status":"completed","duration_ms":10178,
 "details":{"deps_failed":0,"deps_scanned":1,"deps_skipped":0,"deps_with_findings":0,"total_dep_findings":0}}
{"phase":"scan","status":"completed","duration_ms":15352}
```

The report holds the finding the issue names, `DES-ECB-PKCS5Padding` in `services/ledger/src/main/java/app/Use.java`, and `commons-lang3` was resolved by a real `mvn` run from the nested `pom.xml`.

## How

When the scan root holds no manifest for the selected ecosystem, one bounded walk finds the module directories below it, each is resolved by the same ecosystem resolver, and the results merge into the single `ResolveResult` the pipeline already consumes. Each discovered root becomes a `WorkspaceMember`, which is the semantics the field already documents for a monorepo's own modules.

`internal/dependency/roots.go` owns the new knowledge. `rootManifests` is a table of ecosystem to the files that make a directory a resolution root, which replaces another chain of scattered `os.Stat` calls. `ResolutionRoots` is the gate plus the walk, `MergeRootResolutions` folds the results. `internal/engine` owns the fan-out loop, in `resolveScanRoot`, because that is where `failure` wrapping belongs under this repo's two-layer error model.

`prepareDependencyScan` changes one line. `internal/cli`, `internal/dependency/resolver.go` and `collectPackageSets` are untouched.

Maven and Gradle modules in the same repository both resolve, with no Java-specific code, because `JavaResolver` calls `DetectJavaBuildTool` per invocation and dispatches per directory.

## Bounds

A directory qualifies up to four levels below the scan root, at most 16 roots are resolved, and the walk stops at 50000 entries. The root cap bounds process execution, since one root is one build-tool invocation plus, for Maven, a source-jar pass. Resolution is sequential on purpose: concurrent `mvn` processes contend on the same `~/.m2` and the same source cache, and sequential execution makes the caller's deadline a natural global bound.

Build output, dependency, hidden, test and fixture directories are not searched, reusing `skip.DefaultDirMatcher()`. A qualifying root is not descended into, because its own build tool reports the modules it declares; descending would resolve them twice and parse one directory under two import paths.

All three caps are injected through an unexported twin, following the `internal/skip/built_output.go` pattern, so the tests bind them without a shared variable and changing one is a constant, not a design.

## What stays byte-identical

| Scan root | Behavior |
|---|---|
| holds a manifest for the selected ecosystem | unchanged, same errors including `java_build_tool_ambiguous` |
| ecosystem is Python | unchanged, the gate never opens; pip resolves from the interpreter, so fanning out would run the same query per subdirectory for the same answer |
| holds no manifest and nothing qualifies below | unchanged, the same typed error with the same code and details |
| holds no manifest and manifests sit below | resolved, which is this fix |

Deliberately not covered: a root that has a manifest but whose undeclared sibling modules sit below it. Opening the gate there would change behavior for existing users.

## Three decisions worth a reviewer's attention

**`RootModule` is the scan root's directory name, never empty.** A merged result has no module identity of its own, but empty is not available. `exportUserPackages` (`internal/scan/export.go:3319`) returns a nil stop-set for an empty `RootModule`, which makes `liveReachability` report `not_applicable` for every finding, and `occurrenceSourceSubject` (`internal/scan/occurrence_key.go:162`) hashes every non-dependency finding against it. The directory-name fallback is the convention `DetectRootModule` already documents and already applies to this exact tree on the `--export-callgraph` path.

**Dependencies are concatenated, not deduped.** `canonicalDependencies` (`internal/engine/dependency_scanner.go:1070`) already dedupes by `Module@Version`, already prefers the copy that has a `Dir`, and already back-fills the artifact paths. Duplicating that rule in the merge would be information leakage, so a test pins that the engine stays its owner. The practical effect is that a dependency reached from two modules still gets one scan.

**No progress vocabulary change.** ADR 0001 states that a completed `dependencies` phase's `details` contains only its five integer counters, so the discovery counts go to logs. No new phase, status, skip reason or failure code, and no renamed code.

## Tests

18 new tests in `internal/dependency/roots_test.go` covering the gate, the prune rules, all three caps, truncation ordering, an unreadable directory, and every merge rule. 5 new tests in `internal/engine/dependency_scanner_test.go`.

`TestDependencyScanner_ScanWithDependencies_NestedPomWithoutMaven` is the regression proof and needs no `mvn`: it runs a real `JavaResolver` against a nested `pom.xml` with no declared dependencies, with `PATH` set to an empty directory, and Maven's own fast path returns before spawning anything. Reverting the one-line `prepareDependencyScan` change makes it fail with the exact envelope from the issue.

## Verification

- `go build ./...` clean
- `go test ./internal/dependency/... ./internal/engine/... ./internal/scan/...` all ok
- `make lint` with the pinned golangci-lint v2.10.1, `0 issues.`
- the CLI reproduction above, run before and after
- the user guide checked per `docs/user-guide/AGENTS.md`: HTML parsed, prohibited-term grep clean, `git diff --check` clean

`go test ./...` is red in `internal/callgraph/contracts` on `main` itself, independent of this branch, which touches no file in that package. Filed as #534.

## Relationship to #391

Separate, and untouched here. #391 is that a failing resolution aborts the whole scan instead of routing to the existing skip. This branch does not change that policy; a failing root still aborts, but the message now names which root failed. When no root is discovered, the existing error passes through exactly as before.


matiasdaloia:
The red `test` check here is inherited from `main`, not caused by this branch.

The only failure is `TestLoadEmbeddedNodeFastJwt` in `internal/callgraph/contracts`, a package this diff does not touch. It fails the same way on a clean checkout of `main` at `22a9533`. It is issue #534, and #536 fixes it.

Verified: merging #536's branch into this one locally and running `go test ./...` leaves the suite clean. So this PR goes green on its own once #536 lands and this branch picks up `main`. I have not rebased or retargeted anything here.

Everything this branch owns is green. `build` (the pinned golangci-lint) passes, and all 23 new tests pass in CI.

matiasdaloia:
Correction to my note above: the fix for the inherited failure is **#538**, not #536. I closed #536 as a duplicate once I saw #538 covers the same defect and one test file more.

Everything else stands. The only red check here is `TestLoadEmbeddedNodeFastJwt`, inherited from `main`, in a package this diff does not touch. Merging #538's branch into this one locally leaves `go test ./...` clean, so this PR goes green once #538 lands and this branch picks up `main`.

PR #536, opened 2026-09-18
test(contracts): assert library key ownership instead of sibling absence

### LINKED ISSUE #534: main is red: TestLoadEmbeddedNodeFastJwt sees jsonwebtoken keys after the jsonwebtoken KB landed

matiasdaloia:
Closes #534.

## What is broken

`main` ships a red test suite. On a clean checkout of `22a9533`:

```
go test ./internal/callgraph/contracts/ -run TestLoadEmbeddedNodeFastJwt
```

```
--- FAIL: TestLoadEmbeddedNodeFastJwt (0.01s)
    node_fast_jwt_kb_test.go:81: key "jsonwebtoken.sign#3" belongs to jsonwebtoken., a different package
    node_fast_jwt_kb_test.go:81: key "jsonwebtoken.decode#2" belongs to jsonwebtoken., a different package
    ... 6 more
```

No network, no toolchain, pure embedded-YAML loading, so it fails the same way in CI. Every PR opened against `main` inherits it, which is how a genuinely broken PR becomes indistinguishable from a healthy one.

## Why the assertion was never right

The `fast-jwt` test asserted that no key prefixed `jsonwebtoken.` or `jose.` existed in the loaded Node knowledge base. `contracts.LoadEmbedded` discovers every `*.yaml` in the ecosystem directory and merges them, so a sibling library's keys are supposed to be there. The assertion only held while neither package had a knowledge base, and the `jsonwebtoken` KB added in #530 ended that.

The invariant the test actually wanted is ownership. A library keeps to its own prefix, and does not author keys outside it. That is what `assertLibraryOwnsPrefix` states, over `Contract.SourceLibrary`, which `Load` populates from the YAML `library.name` field. It stays correct as further Node libraries land.

## The same mistake was already written twice

`node_elliptic_kb_test.go` named `secp256k1.`, `@noble/curves.` and `tweetnacl.` as packages that must not appear. Each one would have broken that test on the day its KB arrived, for a collision that never happened. Both call sites now use the shared helper, so the next KB author has a correct check to reach for instead of a fresh list of siblings to go stale.

The newest test in this family, `node_jsonwebtoken_kb_test.go`, already got this right by checking that its own method names are not authored under a foreign prefix. This change brings the two older tests up to that standard and makes the check reusable.

## Verification

```
go test ./internal/callgraph/contracts/
ok  	github.com/scanoss/crypto-finder/internal/callgraph/contracts	2.073s
```

Failed with eight errors before the change. The knowledge bases themselves are untouched and no contract key moves, so this changes no scanning behavior. No CHANGELOG entry, since `AGENTS.md` excludes test-only changes.

## Found while

Working on #533. That PR (#535) cannot go green until this lands.


matiasdaloia:
Closing in favour of #538, which fixes the same defect and covers more.

Both PRs were opened within 22 minutes of each other by two sessions working the same red `main`, and both landed on the same mechanism: assert ownership through `Contract.SourceLibrary` rather than asserting a sibling package is absent from the merged knowledge base. #538 additionally updates `node_jsonwebtoken_kb_test.go`, which this PR left alone. Both are green.

Nothing here is lost. #534 stays open until #538 lands.

PR #538, opened 2026-09-18
test(contracts): scope the node KB foreign-coordinate check to its own library

matiasdaloia:
## Why

`main` is red. `TestLoadEmbeddedNodeFastJwt` fails on the unmodified tip `22a9533`, so every open PR inherits a failing `Test & Coverage` job through no fault of its own. Bisected: the parent `88185e8` passes, and #530 (`feat(contracts): add jsonwebtoken Node contract KB`) is the commit that turns it red.

Nothing is wrong with that KB. Each node library's test asks whether that library claims a neighbor's coordinates, which is a real invariant, but answers it by scanning the MERGED knowledge base for the neighbor's key prefix. `LoadEmbedded` merges every YAML in the ecosystem directory, so the scan holds only while the neighbor has no KB of its own, and adding one is the documented way this catalog grows. Adding `jsonwebtoken.yaml` made `jsonwebtoken.` keys appear in the merged KB, and the fast-jwt test reported them as fast-jwt's, although fast-jwt had not changed.

`SourceLibrary` is already populated on every contract from the YAML's `library.name`. Asserting through it answers the question that was meant, and keeps answering it however many neighbors gain a KB.

## Scope

- `assertLibraryOwnsItsKeys` in the new `internal/callgraph/contracts/node_kb_ownership_test.go`: every contract a library contributes is keyed under that library's own prefix.
- `node_fast_jwt_kb_test.go`, `node_elliptic_kb_test.go` and `node_jsonwebtoken_kb_test.go` move to it. Only the first is failing today. The other two carry the same scan against `secp256k1.`, `@noble/curves.`, `tweetnacl.` and `jose.`, each armed to fail the day those KBs land, so they are converted in the same wave rather than left to break one at a time.
- Test-only. No KB, loader or production code changes, so no CHANGELOG entry per AGENTS.md.

## Blast Radius

Unblocks CI for every open PR. The assertion is strictly stronger than the one it replaces: the prefix scan could not see a fast-jwt contract keyed under a package with no KB, and this one can.

## Verification

`go test ./internal/callgraph/contracts/` fails on `origin/main` and passes here.

The new assertion is proved able to fail rather than assumed. Injecting one contract keyed `jsonwebtoken.poisonProbe` into `fast-jwt.yaml` produces `fast-jwt contributes key "jsonwebtoken.poisonProbe#0", outside its own "fast-jwt." coordinates`, and removing it returns the package to green. The KB file is unchanged in this diff.

`make lint` reports 0 issues.


matiasdaloia:
I opened #536 for the same defect before seeing this one, and have closed it in favour of this PR, which covers `node_jsonwebtoken_kb_test.go` as well.

One thing from that branch worth folding in if you agree: `gocritic rangeValCopy` fires on `for _, c := range declared` over `[]Contract`, because the struct is 168 bytes. It failed the `build` job until I switched to indexing the slice. Worth checking this branch's helper does the same, since `make lint` is a required check.

#535 is blocked on this landing, and is otherwise green.

